PR #6, opened 2018-08-28
WIP: API, CSV backend, dependencies and more

tebeka:
The PR is mostly to align code for working in parallel.

Known issues:
- Bug in server write - doesn't report last chunk
- Bug in server read - doesn't serve last package
- Go client write support untested


backends/utils/utils.go
```diff
@@ -0,0 +1,116 @@
...
+	return nil, fmt.Errorf("unknown type - %T", value)
+}
+
+// AppendNil appends an empty value to data
+func AppendNil(col frames.Column) error {
+	switch col.DType() {
+	case frames.IntType:
+		return col.Append(0)
+	case frames.FloatType:
+		return col.Append(0.0)
```
yaronha: @tebeka should be math.NaN() ?

mapframe.go
```diff
@@ -0,0 +1,172 @@
...
+	return NewMapFrame(frameSlice)
+}
+
+// MapFrameMessage is over-the-wire frame data
+type MapFrameMessage struct {
+	Columns   []string                       `msgpack:"columns"`
+	SliceCols map[string]*SliceColumnMessage `msgpack:"slice_cols,omitempty"`
+	LabelCols map[string]*LabelColumnMessage `msgpack:"label_cols,omitempty"`
+}
+
```
yaronha: @tebeka looks like some redundancy : 
1. Col names are in the col object 
2. Slice & Label can be part of the same list just different data types (Interface) 

also we need an indication which col is an index col (optional)

server/server.go
```diff
@@ -0,0 +1,325 @@
...
+			ctx.Error(err.Error(), http.StatusBadRequest)
+			return
+		}
+	}
+
+	// TODO: Validate request
+
+	// TODO: We'd like to have a map of name->config of backends in configuration
+	backend, err := s.createBackend(request.Type)
+	if err != nil {
```
yaronha: @tebeka why do we create the backend every request ? we should create the backend on init

backends/csv/backend.go
```diff
@@ -0,0 +1,343 @@
...
+		return nil, err
+	}
+
+	ca := &csvAppender{
+		writer:    file,
+		csvWriter: csv.NewWriter(file),
+		logger:    b.ctx.Logger,
+	}
+
+	return ca, nil
```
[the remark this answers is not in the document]
tebeka: Yup. Fixed.

backends/utils/utils.go
tebeka: Done

server/server.go
tebeka: Yeah. We talked about it and it - #7

PR #100, opened 2018-11-18
Exec command

tebeka:


yaronha:
@tebeka need to add expression (string) as another param in exec

PR #148, opened 2018-12-24
update v3io-go, add support for session keys, align env variables

yaronha:


clients/py/v3io_frames/__init__.py
```diff
@@ -67,9 +67,9 @@ def Client(address='', data_url='', container='', path='', user='',
         url=data_url or env.url,
         container=container or env.container,
         path=path or env.path,
-        user=user or env.user,
-        password=password or env.password,
-        token=token or env.token,
+        user=user or env.user or environ.get('V3IO_USERNAME'),
```
tebeka: I'm not sure that having  both `V3IO_SESSION` and `V3IO_USERNAME` is a good idea. If we do decide to keep both then:
1. This code need to move to `session_from_env`
2. We should probably look for all `V3IO_*` environment variables and align them with session field (e.g. `V3IO_USER` instead of `V3IO_USERNAME`)
3. Need to be clear what has higher priority. `V3IO_SESSION` or a specific field environment variable.

backends/stream/backend.go
```diff
@@ -127,19 +127,14 @@ func (b *Backend) Exec(request *frames.ExecRequest) error {
...
-		session.User, session.Password,
 		b.backendConfig.Workers,
 	)
 
-	if err != nil {
-		return nil, errors.Wrap(err, "Failed to create V3IO data container")
-	}
-
-	return container, nil
+	return container, err
```
tebeka: Why? I prefer the former version, it's more explicit that there's no error.

clients/py/v3io_frames/client.py
```diff
@@ -34,7 +34,7 @@ def __init__(self, address, session):
         session : Session
             Session object
         """
-        address = address or environ.get('V3IO_FRAMES_ADDR')
+        address = address or environ.get('V3IO_FRAMESD')
```
tebeka: I prefer to leave the `_ADDR` prefix.

clients/py/v3io_frames/__init__.py
yaronha: @tebeka agree we dont need both, didnt want to break the tests. the V3IO_** are now aligned w all other V3IO based environments and our cloud provisioning which is why i changed some of them

backends/stream/backend.go
yaronha: @tebeka i moved some stuff to common place (will also help when we use the session UID to cache v3io container objects..), its ok to change specifically this err thing if u want

clients/py/v3io_frames/client.py
yaronha: @tebeka its the new cloud convention

vendor/github.com/v3io/v3io-go-http/context.go
```diff
@@ -11,6 +11,13 @@ type Context struct {
 	numWorkers  int
```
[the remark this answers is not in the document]
yaronha: @tebeka yes, now in master

PR #155, opened 2018-12-26
Simplejson support

aghid:
@tebeka , please review this PR


simplejson/simplejson.go
```diff
@@ -0,0 +1,268 @@
...
+type TimeSeriesOutput struct {
+	Datapoints [][]interface{} `json:"datapoints"`
+	Target     string          `json:"target"`
+}
+
+const QUERY_SEPARATOR = ";"
+const FIELDS_ITEMS_SEPARATOR = ","
+const DEFAULT_BACKEND = "tsdb"
+
+func SimpleJsonRequestFactory(method string, request []byte) (SimpleJsonRequestInterface, error) {
```
tebeka: IMO this is not required since in the handler you know which one you need.

simplejson/simplejson.go
```diff
@@ -0,0 +1,268 @@
...
+		return nil, fmt.Errorf("Unknown method, %s", method)
+	}
+
+	if err := retval.ParseRequest(request); err != nil {
+		return nil, err
+	}
+	return retval, nil
+}
+
+func (req *SimpleJsonQueryRequest) GetReadRequest() *frames.ReadRequest {
```
tebeka: Indent one field per line.

simplejson/simplejson.go
```diff
@@ -0,0 +1,268 @@
...
+		}
+	}
+	if iter.Err() != nil {
+		return nil, iter.Err()
+	}
+	return retval, nil
+}
+
+func (req *SimpleJsonQueryRequest) extractColumnValuesTSDB(column *pb.Column) (reflect.Value, string) {
+	colType := strings.ToLower(column.Dtype.String())
```
tebeka: Why lower case?

simplejson/simplejson.go
```diff
@@ -0,0 +1,268 @@
...
+	dataTypes := strings.Title(fmt.Sprintf("%ss", colType))
```
tebeka: I'd rather you'd do a `switch` on `column.DType`, this is relying on implementation details of column that might change.

simplejson/simplejson.go
```diff
@@ -0,0 +1,268 @@
...
+		if err := req.parseQueryLine(strings.TrimSpace(target)); err != nil {
+			return err
+		}
+	}
+	return nil
+}
+
+func (req *SimpleJsonQueryRequest) parseQueryLine(fieldInput string) error {
+	translate := map[string]string{"table_name": "TableName"}
+	re := regexp.MustCompile(`^\s*(filter|fields|table_name|backend|container|username|password)\s*=\s*(.*)\s*$`)
```
tebeka: Please an example of input as comment here

simplejson/simplejson.go
```diff
@@ -0,0 +1,268 @@
+package simplejson
+
+import (
+	"encoding/json"
+	"errors"
+	"fmt"
+	"reflect"
```
tebeka: IMO too much use of reflection. Can be done without it.

simplejson/simplejson.go
tebeka: This should probably be in `init` so we'll crash before we start the server if the regexp is not valid. (fail fast)

simplejson/simplejson.go
```diff
@@ -0,0 +1,268 @@
...
+			fieldName := strings.Title(match[1])
+			if fieldNameTranslated, ok := translate[match[1]]; ok {
+				fieldName = fieldNameTranslated
+			}
+			if fieldName == "Fields" {
+				value = strings.Split(match[2], FIELDS_ITEMS_SEPARATOR)
+			} else {
+				value = match[2]
+			}
+			if err := setField(req, fieldName, value); err != nil {
```
tebeka: Not very idiomatic Go :) 
Can you give some example of query lines?

http/server.go
```diff
@@ -399,15 +352,79 @@ func (s *Server) handleExec(ctx *fasthttp.RequestCtx) {
...
+func (s *Server) handleSimpleJsonQuery(ctx *fasthttp.RequestCtx) {
+	req, err := simplejson.SimpleJsonRequestFactory("query", ctx.PostBody())
+	if err != nil {
+		s.logger.ErrorWith("can't initialize simplejson query request", "error", err)
+		ctx.Error(fmt.Sprintf("bad request - %s", err), http.StatusBadRequest)
+		return
+	}
+	readRequest := req.GetReadRequest()
+	factory := backends.GetFactory(readRequest.Backend)
+	var backendConfig *frames.BackendConfig
```
tebeka: Use `s.api.Read` instead of manually creating a backend here.

aghid:
Hi @tebeka, regarding the context, we want to add an interface for grafana simplejson: https://github.com/grafana/simple-json-datasource (the documentation contains examples of requests and responses)

http/server.go
```diff
@@ -399,15 +352,79 @@ func (s *Server) handleExec(ctx *fasthttp.RequestCtx) {
...
-		"/write":    s.handleWrite,
-		"/exec":     s.handleExec,
+		"/_/config":                  s.handleConfig,
+		"/_/status":                  s.handleStatus,
+		"/create":                    s.handleCreate,
+		"/delete":                    s.handleDelete,
+		"/read":                      s.handleRead,
+		"/write":                     s.handleWrite,
+		"/exec":                      s.handleExec,
+		"/grafana/simplejson/":       s.handleSimpleJsonACK,
```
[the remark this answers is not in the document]
aghid: @tebeka , we may use other grafana plugins in the future and the idea was to have them under "/grafana/<plugin>". I will consult with @yaronha to see if this is indeed something that he wants. Thanks!

simplejson/simplejson.go
aghid: This (query) is part of the request payload

simplejson/simplejson.go
aghid: I know, @tebeka , GO is not my "mother tongue", my background is python so it felt very natural to me to have a "setattr" LOL. The idea was to have a generic way to set the request fields once parsed from the query. Here is an example of such a query: `fields=sentiment;table_name=stock_metrics;backend=tsdb;filter=symbol==\"AAPL\";container=containerName;username=someuser;password=xyz`

http/server.go
aghid: I had a discussion with @yaronha on this and he suggested using backend.Read instead of api.Read to avoid using a channel

simplejson/simplejson.go
```diff
@@ -0,0 +1,268 @@
...
+	SimpleJsonQueryRequest
+}
+
+type TableOutput struct {
+	Columns []map[string]string `json:"columns"`
+	Rows    [][]interface{}     `json:"rows"`
+	Type    string              `json:"type"`
+}
+
+type TimeSeriesOutput struct {
```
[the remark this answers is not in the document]
aghid: Indeed I am :)

simplejson/simplejson.go
aghid: it's one way to have the read function get called always. I will make all these requests non-exported so the factory will be required. What say you?

simplejson/simplejson.go
aghid: I'm using reflect when this might help with performance (passing reflect.Value instead of "converting" a slice of a certain type to interface slice) but also to add some generic implementation, knowing that GO is not quite generic friendly. The latter is used only to set the struct attributes and can be quickly replaced with switch/case.

simplejson/simplejson.go
tebeka: I mean the `re.MustCompile` which might panic on invalid regexp

http/server.go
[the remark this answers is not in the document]
tebeka: Talk to @yaronha . He perers just `/query` and `/search` at the top level. I wonder if we can use the existing `/read` for `query` with maybe a `?format=grafana`.

simplejson/simplejson.go
```diff
@@ -0,0 +1,268 @@
...
+	Type      string
+	Backend   string
+	From      string
+	To        string
+	Container string
+	Username  string
+	Password  string
+}
+
+type SimpleJsonSearchRequest struct {
```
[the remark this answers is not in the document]
tebeka: Since it's a `1 to 1` between query and response, I don't see the value of generalization here.

simplejson/simplejson.go
tebeka: Don't see the value of not exporting these. I prefer simpler code every time :)

simplejson/simplejson.go
tebeka: Have a look at https://github.com/mitchellh/mapstructure (we use it in another project). I think it's better to have parsers return a map and then use `mapstructure` to populate the structure.

simplejson/simplejson.go
tebeka: Yeah, coming from Python myself it takes a bit to get used to writing idiomatic Go. 
The input example should be a comment in the code - makes it easier to understand the parser.

Something like:
```go
// Parse a query line
// query example:
// fields=sentiment;table_name=stock_metrics;backend=tsdb;filter=symbol==\"AAPL\";container=containerName;username=someuser;password=xyz`
func parseQuery(queryLine string) (map[string]string, error) {
	query := make(map[string]string)
	for fNum, field := range strings.Split(queryLine, ";") {
		i := strings.Index(field, "=")
		if i == -1 {
			return nil, fmt.Errorf("%d: bad field = %s", fNum, field)
		}
		query[field[:i]] = field[i+1:]
	}

	return query, nil
}
```

http/server.go
tebeka: Why should we avoid using a channel?

yaronha: @tebeka the channel was from the http to the API (since we used streaming) AFAIK, here we dont have a front-end chan and no need for one

http/grafana.go
```diff
@@ -0,0 +1,288 @@
...
+	for iter.Next() {
+		frame := iter.At()
+		iface, ok := frame.(pb.Framed)
+		if !ok {
+			return nil, errors.New("unknown frame type")
+		}
+		for _, column := range iface.Proto().Columns {
+			values, colType := req.extractColumnValues(column)
+			retVal.Columns = append(retVal.Columns, Column{Text: column.Name, Type: colType})
+			for rowIndex := 0; rowIndex < values.Len(); rowIndex++ {
```
tebeka: Why not use `frame.IterRows`?

http/grafana.go
```diff
@@ -0,0 +1,288 @@
...
+		}
+	}
+	if iter.Err() != nil {
+		return nil, iter.Err()
+	}
+	return retval, nil
+}
+
+func (req *simpleJsonQueryRequest) extractColumnValues(column *pb.Column) (reflect.Value, string) {
+	colTypeStr := strings.ToLower(column.Dtype.String())
```
tebeka: Use extranal frame API here, the `Column` interface implements all that you want.

http/grafana.go
```diff
@@ -0,0 +1,288 @@
...
+		frame := iter.At()
+		iface, ok := frame.(pb.Framed)
+		if !ok {
+			return nil, errors.New("unknown frame type")
+		}
+		for _, column := range iface.Proto().Columns {
+			values, colType := req.extractColumnValues(column)
+			retVal.Columns = append(retVal.Columns, Column{Text: column.Name, Type: colType})
+			for rowIndex := 0; rowIndex < values.Len(); rowIndex++ {
+				if len(retVal.Rows) <= rowIndex {
```
tebeka: This should  be outside the current loop - just once.

http/grafana.go
```diff
@@ -0,0 +1,288 @@
...
+		break
+	case pb.DType_BOOLEAN:
+		dataSlice = &column.Bools
+		break
+	}
+	return reflect.ValueOf(dataSlice).Elem(), colTypeStr
+}
+
+func (req *simpleJsonQueryRequest) formatTimeTSDB(timestamp time.Time) interface{} {
+	return timestamp.UnixNano() / 1000000
```
tebeka: nitpicking: `/ 1e6` is more readable IMO

http/grafana.go
```diff
@@ -0,0 +1,288 @@
...
+			} else {
+				labels += fmt.Sprintf("[%s=%s]", column.Name, val)
+			}
+		}
+	}
+	return metricName + labels
+}
+
+func (req *simpleJsonQueryRequest) CreateResponse(iter frames.FrameIterator) (interface{}, error) {
+	formatters := map[string]func(iter frames.FrameIterator) (interface{}, error){"kv": req.formatKV, "tsdb": req.formatTSDB}
```
tebeka: Do a switch here, we're not in Python :)

http/grafana.go
```diff
@@ -0,0 +1,288 @@
...
+	if formatter, ok := formatters[req.Backend]; ok {
+		return formatter(iter)
+	}
+	return nil, fmt.Errorf("Unknown format: %s", req.Backend)
+}
+
+func (req *simpleJsonSearchRequest) CreateResponse(iter frames.FrameIterator) (interface{}, error) {
```
tebeka: This is one reaon why I don't like the generic approach with `simpleJsonRequestInterface`. It is used just on one place and here it's not supported - [YAGNI](https://martinfowler.com/bliki/Yagni.html)

http/server.go
```diff
@@ -399,15 +353,101 @@ func (s *Server) handleExec(ctx *fasthttp.RequestCtx) {
 	s.replyOK(ctx)
 }
 
+func (s *Server) handleSimpleJsonQuery(ctx *fasthttp.RequestCtx) {
+	username, password := s.extractBasicAuth(ctx)
```
tebeka: See [here](https://github.com/v3io/frames/blob/development/http/server.go#L412), was added lately.
Also Go common practice is to call it `handleSimpleJSONQuery` (JSON)

http/server.go
tebeka: After talking to @yaronha a bit more, it's better to use the API and not create backend every time, here's a code skeleton
```go
	ch := make(chan frames.Frame)
	var apiError error
	go func() {
		defer close(ch)
		apiError = s.api.Read(request, ch)
		if apiError != nil {
			s.logger.ErrorWith("error reading", "error", apiError)
		}
	}()

	for frame := range ch {
		// Do stuff here
	}

	if apiError != nil {
		s.writeError(enc, apiError)
	}
```

http/server.go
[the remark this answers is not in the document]
aghid: grafana (the main client for this feature) appends `/query` to the url set in the datasource so `/read` will not work - already made the changes to the paths.

simplejson/simplejson.go
aghid: will check it out, thanks

simplejson/simplejson.go
aghid: worst error I got is "failed to create adapter with a random string, but I have replaced it with "Compile" for better error handling

simplejson/simplejson.go
aghid: thanks for that, but I still need to inject the query into the simple json request, "Backend", "Fields".....

PR #188, opened 2019-01-29
ColumnBuilder.Delete

tebeka:


yaronha:
@tebeka I believe we need a range (delete from i to j)

tebeka:
@yaronha Talked to @talIguaz - I don't think range delete is a common case. And I rather keep the API simple - people can use a `for` loop for deleting a range.

PR #195, opened 2019-02-17
Fix cuda catgorical concat

tebeka:


tebeka:
Fixes #190

tebeka:
See [here](https://github.com/v3io/frames/blob/development/_scripts/install-cudf-env.sh) for current install on `NVIDIA Volta Deep Learning AMI`

PR #208, opened 2019-03-11
Drop support for vendor dir.

gtopper:


go.mod
```diff
@@ -26,7 +26,6 @@ require (
 	go.uber.org/multierr v1.1.0 // indirect
 	golang.org/x/net v0.0.0-20181023162649-9b4f9f5ad519
 	golang.org/x/sys v0.0.0-20181026203630-95b1ffbd15a5 // indirect
-	google.golang.org/appengine v1.2.0 // indirect
```
tebeka: How are these changes and the onces on `go.sum` connected to this PR?

gtopper: Basically I just ran `go mod tidy` and committed the result. 🙂

gtopper:
For patching a 3rd party dependency, the new way it to use `replace` e.g. https://github.com/v3io/v3io-tsdb/blob/development/go.mod#L32, which allows you to use a proper fork.

As to internet access, I think it's not supposed to be a problem if you run `go mod tidy` after cloning the main repository.

PR #223, opened 2019-04-21
IG-11257: Add boolean support.

gtopper:


gshatz:
@gtopper please add test for booleans

gtopper:
There are no integration tests in frames. There is already a unit test checking booleans (which I guess means it doesn't test much). So adding a proper test for this is likely to take a significant effort.

PR #227, opened 2019-05-08
IG-11266: Support writing and reading dataframes with unnamed columns.

gtopper:
🐍

### DIFF
--- a/clients/py/v3io_frames/pbutils.py
+++ b/clients/py/v3io_frames/pbutils.py
@@ -53,12 +53,12 @@ def SchemaField(name=None, doc=None, default=None, type=None, properties=None):
     """A schema field"""
     # We return a frames_pb2.SchemaField from Python types
     return fpb.SchemaField(
-            name=name,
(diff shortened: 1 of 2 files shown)


clients/py/v3io_frames/pbutils.py
```diff
@@ -141,18 +157,25 @@ def df2msg(df, labels=None, index_cols=None):
         else:
             serieses = [idx2series(df.index)]
 
-        indices = [series2col(s) for s in serieses]
+        indices = [series2col(s, s.name) for s in serieses]
+
+    columns = []
+    for name in df.columns:
+        series = df[name]
+        if isinstance(name, int):
```
talIguaz: isn't for every column named as an int you will prefix it with `column_`?
and then when you read the DF you will get columns: `column_23, column_5` instead of `23,5`.
is that what we want?

gtopper: Added a check.

PR #230, opened 2019-05-15
TSDB-116: Fix propagation of password and token in TSDB create.

gtopper:


gtopper:
Closing, because we won't make a new release for this.

PR #235, opened 2019-05-26
Fix HTTP server session handling.

gtopper:


gtopper:
Reopening with serialization fixes.

PR #240, opened 2019-06-06
IG-12266: DeleteTable rewrite.

gtopper:


v3ioutils/container.go
```diff
@@ -155,48 +162,50 @@ func DeleteTable(logger logger.Logger, container v3io.Container, path, filter st
...
+}
 
-			case <-time.After(timeout):
-				if !active {
-					logger.ErrorWith("Resp loop timed out!", "requests", requests, "response", responses)
-					done <- true
-					return
-				}
-				active = false
+func deleteObjectWorker(tablePath string, container v3io.Container, fileNameChan <-chan string, outgoingTerminationChan chan<- error, onErrorTerminationChannel <-chan struct{}) {
```
talIguaz: i would change `outgoingTerminationChan` to just `terminationChan` so it would be the same as the getItemsWorker for clarity

PR #249, opened 2019-07-03
IG-12574

talIguaz:
* write index col as attribute
* do not cast numeric columns to floats

### DIFF
--- a/backends/kv/writer.go
+++ b/backends/kv/writer.go
@@ -272,7 +272,6 @@ func (a *Appender) WaitForComplete(timeout time.Duration) error {
 
 func (a *Appender) indexValFunc(frame frames.Frame) (func(int) interface{}, error) {
 	var indexCol frames.Column
-	var fn func(int) interface{}
 
 	if indices := frame.Indices(); len(indices) > 0 {
 		if len(indices) != 1 {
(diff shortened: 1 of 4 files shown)


backends/kv/writer.go
```diff
@@ -259,8 +270,9 @@ func (a *Appender) WaitForComplete(timeout time.Duration) error {
 	return a.asyncErr
 }
 
-func (a *Appender) indexValFunc(frame frames.Frame) (func(int) string, error) {
+func (a *Appender) indexValFunc(frame frames.Frame) (func(int) interface{}, error) {
 	var indexCol frames.Column
+	var fn func(int) interface{}
```
gtopper: Why move this?

talIguaz: by mistake, ill change it back

cmd/framesd/framesd.go
```diff
@@ -49,7 +49,7 @@ func main() {
 		showVersion bool
 	}
 
-	flag.StringVar(&config.file, "config", "", "path to configuration file (YAML)")
+	flag.StringVar(&config.file, "config", "cfg.yaml", "path to configuration file (YAML)")
```
gtopper: ⚠️

PR #254, opened 2019-07-10
Arrow - WIP

tebeka:
Initial implementation of using arrow. This implements only CSV backend write to plasma shared memory store and Python client reading from it. Currently on gRPC API changes to have `ShmRead` method.

See [here](https://docs.google.com/document/d/1dVtt-79C0Of5stJy9WutC4OIgMs3CNgvR360roDrKbk/edit#heading=h.1hcibne0a81) on how to run the code on EC2

@bivas Please take a look at `cmd/frames/Dockerfile.arrow` and let's discuss it.


arrow.go
```diff
@@ -0,0 +1,467 @@
...
+	return fmt.Errorf("not implemented")
+}
+
+// Delete deletes a value
+func (b *ArrowColumnBuilder) Delete(index int) error {
+	return fmt.Errorf("not implemented")
+}
+
+// Finish create the colum
+func (b *ArrowColumnBuilder) Finish() Column {
```
bivas: Usually this is called `Build` or `Get`

grpc/arrow.go
```diff
@@ -0,0 +1,150 @@
...
+
+	var apiError error
+	go func() {
+		apiError = s.api.Read(r, ch)
+		if apiError != nil {
+			s.logger.ErrorWith("API error reading", "error", apiError)
+		}
+	}()
+
+	frame := <-ch // TODO: timeout
```
bivas: or use the context you get as argument.

arrow.go
tebeka: Agree. However I'm trying to keep the same name [Arrow C++ uses](https://arrow.apache.org/docs/cpp/classarrow_1_1_array_builder.html#a67b2227edb4b51bc44425f1c8a439819)

grpc/arrow.go
tebeka: Agree, very initial stages here :)

PR #260, opened 2019-07-16
IG-12726: Fix panic when reading KV with specific columns containing index

talIguaz:
if a user requested specific columns on kv read, and one of them is the index col, duplicate the index column to be as both index and regular column. the index name would be  `_<indexname>`

### DIFF
--- a/backends/kv/reader.go
+++ b/backends/kv/reader.go
@@ -23,6 +23,7 @@ package kv
 import (
 	"encoding/json"
 	"fmt"
+
 	"github.com/v3io/frames"
 	"github.com/v3io/frames/backends"
 	"github.com/v3io/frames/backends/utils"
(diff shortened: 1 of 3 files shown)


backends/kv/reader.go
```diff
@@ -22,6 +22,7 @@ package kv
 
 import (
 	"encoding/json"
+	"fmt"
```
gtopper: Missing newline for linter 😈

column.go
```diff
@@ -319,6 +319,16 @@ func (c *colImpl) Append(value interface{}) error {
 	return c.appendSlice(value)
 }
 
+func (c *colImpl) CopyWithName(newName string) Column {
+	var newCol Column
+	newColImpl := *c
+	newCol = &newColImpl
```
gtopper: Inline?

PR #262, opened 2019-07-16
IG-12459 - KV support reading partitioned table

talIguaz:


backends/kv/reader.go
```diff
@@ -212,6 +220,65 @@ func init() {
...
+		if err != nil {
+			return nil, err
+		}
+		partitions = append(partitions, parts...)
+	}
+
+	return partitions, nil
+}
+
+func (kv *Backend) internalGetPartitions(path string, container v3io.Container, marker string) ([]string, error) {
```
gtopper: This is not used 😮

backends/kv/reader.go
```diff
@@ -212,6 +220,65 @@ func init() {
...
+	out := res.Output.(*v3io.GetContainerContentsOutput)
+	for _, partition := range out.CommonPrefixes {
+		parts, err := kv.getPartitions(partition.Prefix, container, "")
+		if err != nil {
+			return nil, err
+		}
+		partitions = append(partitions, parts...)
+	}
+
+	if out.NextMarker != "" {
```
gtopper: You need to use `IsTruncated` for this.

backends/kv/reader.go
```diff
@@ -212,6 +220,65 @@ func init() {
...
+		parts, err := kv.getPartitions(path, container, out.NextMarker)
```
gtopper: You should swap this recursion out for a loop, to avoid a potential stack overflow when a table is very large.

gtopper: Or rather, when it has many partitions.

backends/kv/reader.go
```diff
@@ -212,6 +220,65 @@ func init() {
...
+			partitions = append(partitions, partition.Prefix)
+			parts, err := kv.getPartitions(partition.Prefix, container, "")
+			if err != nil {
+				return nil, err
+			}
+			partitions = append(partitions, parts...)
+		}
+	}
+
+	if out.NextMarker != "" {
```
gtopper: You need to use IsTruncated for this.

backends/kv/reader.go
```diff
@@ -212,6 +220,65 @@ func init() {
...
+		parts, err := kv.getPartitions(path, container, out.NextMarker)
```
gtopper: You should swap this recursion out for a loop, to avoid a potential stack overflow when a table has many partitions.

backends/kv/reader.go
```diff
@@ -212,6 +220,36 @@ func init() {
...
+	var partitions []string
+	input := &v3io.GetContainerContentsInput{Path: path, DirectoriesOnly: true, Marker: marker}
+	res, err := container.GetContainerContentsSync(input)
+	if err != nil {
+		return nil, err
+	}
+	out := res.Output.(*v3io.GetContainerContentsOutput)
+	if len(out.CommonPrefixes) > 0 {
+		for _, partition := range out.CommonPrefixes {
+			partitions = append(partitions, partition.Prefix)
```
gtopper: Looks like this will also include non-lead directories in the partition list.

backends/kv/reader.go
```diff
@@ -52,11 +52,16 @@ func (kv *Backend) Read(request *frames.ReadRequest) (frames.FrameIterator, erro
 		return nil, err
 	}
 
-	input := v3io.GetItemsInput{Path: tablePath, Filter: request.Proto.Filter, AttributeNames: columns}
+	partitions, err := kv.getPartitions(tablePath, container, "")
+	if err != nil {
+		return nil, err
+	}
+
+	input := v3io.GetItemsInput{Path: "", Filter: request.Proto.Filter, AttributeNames: columns}
```
gtopper: `Path: ""` is redundant 😿

backends/kv/reader.go
```diff
@@ -212,6 +217,47 @@ func init() {
...
+			if err != nil {
+				return nil, err
+			}
+			partitions = append(partitions, parts...)
+		}
+	} else { // Add a partition to the list if this is a leaf folder
+		partitions = append(partitions, path)
+	}
+
+	for out.IsTruncated {
```
gtopper: It looks like the whole body of this is duplicated from the above?

backends/kv/reader.go
```diff
@@ -212,6 +217,47 @@ func init() {
 	}
 }
 
+func (kv *Backend) getPartitions(path string, container v3io.Container, marker string) ([]string, error) {
```
gtopper: `marker` is no longer needed - always passed the empty string.

backends/kv/reader.go
gtopper: I'm thinking something along the lines of:
```go
func (kv *Backend) getPartitions(path string, container v3io.Container) ([]string, error) {
	var partitions []string
	var done bool
	var marker string
	for !done {
		input := &v3io.GetContainerContentsInput{Path: path, DirectoriesOnly: true, Marker: marker}
		res, err := container.GetContainerContentsSync(input)
		if err != nil {
			return nil, err
		}
		out := res.Output.(*v3io.GetContainerContentsOutput)
		if len(out.CommonPrefixes) > 0 {
			for _, partition := range out.CommonPrefixes {
				parts, err := kv.getPartitions(partition.Prefix, container)
				if err != nil {
					return nil, err
				}
				partitions = append(partitions, parts...)
			}
		} else if marker == "" { // Add a partition to the list if this is a leaf folder
			partitions = append(partitions, path)
		}
		marker = out.NextMarker
		done = !out.IsTruncated
	}

	return partitions, nil
}
```

backends/kv/reader.go
```diff
@@ -212,6 +217,47 @@ func init() {
...
+	var partitions []string
+	input := &v3io.GetContainerContentsInput{Path: path, DirectoriesOnly: true, Marker: marker}
+	res, err := container.GetContainerContentsSync(input)
```
gtopper: Sadly, you have to call `res.Release()` ❗️

PR #265, opened 2019-07-22
update README

talIguaz:


README.md
```diff
@@ -86,18 +86,13 @@ v3c.write(backend='kv', table='mytable', dfs=tsdf, expression='city="NY"', condi
...
-* group_by: string - Query group by (can't be used with query)
 * columns: []str - List of columns to pass (can't be used with query)
-* limit: int - Maximal number of rows to return
-* row_layout: bool - Weather to use row layout (vs the default column layout)
-* max_in_message: int - Maximal number of rows per message
-* reset_index: bool - Reset the index. When set to `false` (default), the dataframe will have the key column of the v3io kv as the index column.
-When set to `true`, the index will be reset to a range index.
 * data_format: string - Data format (Not yet supported)
 * marker: string - Query marker (Not yet supported)
+* limit: int - Maximal number of rows to return (Not yet supported)
```
gshatz: why do we need all these "not supported" features? why not just delete those options?

talIguaz: Well honestly i left it for me(/us) so that we will know what is half baked in the project.
We can delete the readme entries but then i would like to delete all of the references from the code as well.

PR #274, opened 2019-08-07
IG-12559 && IG-12671

talIguaz:
support v3io timestamps in both read and write of kv.

depends on `v3io-go` support, implemented in - https://github.com/v3io/v3io-go/pull/26


backends/kv/reader.go
```diff
@@ -229,10 +229,12 @@ func (kv *Backend) getPartitions(path string, container v3io.Container) ([]strin
 	for !done {
 		input := &v3io.GetContainerContentsInput{Path: path, DirectoriesOnly: true, Marker: marker}
 		res, err := container.GetContainerContentsSync(input)
-		res.Release() // Releasing underlying fasthttp response
 		if err != nil {
 			return nil, err
 		}
+		if res != nil {
```
gtopper: No need to check this - they can't both be `nil`.

PR #276, opened 2019-08-12
IG-12731

talIguaz:
move kv frame validation to beginning of write


backends/kv/writer.go
```diff
@@ -83,6 +83,9 @@ func (a *Appender) Add(frame frames.Frame) error {
 	if len(names) == 0 {
 		return fmt.Errorf("empty frame")
 	}
+	if len(frame.Indices()) != 1 {
```
gtopper: Can't it also be of length zero?

talIguaz: No. you got to have an index

PR #279, opened 2019-08-21
IG-12938

talIguaz:
handle false false as info message.
Currently engine changed their error formatting and it's from the following format:
```
{
    "ErrorCode": -184549378,
    "ErrorMessage": "UpdateItem: ConditionExpression result is false. Item was not updated",
    "ExtraInfo": {
        "error_0": {
            "ErrorCode": -385875972,
            "ErrorMessage": "Expression error: Attribute doesn't exist"
        },
        "error_1": {
            "ErrorCode": -385876025,
            "ErrorMessage": "Expression error: Key not found for condition"
        }
    }
}
```

Basically we want to ignore only this error -  `ErrorMessage": "Expression error: Key not found for condition`

Later we would need to also change `v3io-go` to support the "new" engine error format

### DIFF
--- a/backends/kv/writer.go
+++ b/backends/kv/writer.go
@@ -22,11 +22,10 @@ package kv
 
 import (
 	"fmt"
-	"strings"
-	"time"
-
 	"github.com/nuclio/logger"
 	v3io "github.com/v3io/v3io-go/pkg/dataplane"
+	"strings"
+	"time"
 
 	"github.com/v3io/frames"
 	"github.com/v3io/frames/backends/utils"
@@ -47,6 +46,12 @@ type Appender struct {
 	asyncErr      error
 	rowsProcessed int
 }
+
+const (
+	errorCodeString              = "ErrorCode"
+	falseConditionOuterErrorCode = "184549378"
+	falseConditionInnerErrorCode = "385876025"
+)
 
 // Write support writing to backend
 func (kv *Backend) Write(request *frames.WriteRequest) (frames.FrameAppender, error) {
@@ -354,8 +359,13 @@ func (a *Appender) respWaitLoop(timeout time.Duration) {
 			timer.Reset(timeout)
 
 			if resp.Error != nil {
-				a.logger.ErrorWith("failed write response", "error", resp.Error)
-				a.asyncErr = resp.Error
+				// If condition was evaluated as false log this and discard error.
+				if isFalseConditionError(resp.Error) {
+					a.logger.Info("condition was evaluated to false for item %v", resp.Request().Input)
+				} else {
+					a.logger.ErrorWith("failed write response", "error", resp.Error)
+					a.asyncErr = resp.Error
+				}
 			}
 
 			if requests == responses {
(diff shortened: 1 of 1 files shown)


backends/kv/writer.go
```diff
@@ -26,11 +26,10 @@ import (
 	"time"
 
 	"github.com/nuclio/logger"
-	v3io "github.com/v3io/v3io-go/pkg/dataplane"
-
 	"github.com/v3io/frames"
 	"github.com/v3io/frames/backends/utils"
 	"github.com/v3io/frames/v3ioutils"
+	v3io "github.com/v3io/v3io-go/pkg/dataplane"
```
gtopper: Pretty sure it was in the right place before.

backends/kv/writer.go
```diff
@@ -425,3 +435,16 @@ func (a *Appender) validateSchema(frame frames.Frame) error {
...
 	return nil
 }
+
+// Check if the current error was caused specifically because the condition was evaluated to false.
+func isFalseConditionError(err error) bool {
+	errString := err.Error()
+
+	if strings.Count(errString, errorCodeString) == 2 &&
+		strings.Contains(errString, falseConditionOuterErrorCode) &&
+		strings.Contains(errString, falseConditionInnerErrorCode) {
```
gtopper: You can use `errors.Cause(err)`.

backends/kv/writer.go
```diff
@@ -354,8 +359,13 @@ func (a *Appender) respWaitLoop(timeout time.Duration) {
 			timer.Reset(timeout)
 
 			if resp.Error != nil {
-				a.logger.ErrorWith("failed write response", "error", resp.Error)
-				a.asyncErr = resp.Error
+				// If condition was evaluated as false log this and discard error.
+				if isFalseConditionError(resp.Error) {
+					a.logger.Info("condition was evaluated to false for item %v", resp.Request().Input)
```
gtopper: We're going to log this every time the condition is false?

backends/kv/writer.go
talIguaz: the GoLand formatter did this 🤷‍♂

backends/kv/writer.go
talIguaz: Not sure that will help since it's just one error with a "big" string like the one in the description.

backends/kv/writer.go
talIguaz: Yes. That's what i get from the ticket.. i am open to changing it to `debug` to not potentially spam the logs but whatever

backends/kv/writer.go
gtopper: `err` is actually a struct. `err.Error()` returns one string.

backends/kv/writer.go
gtopper: Will leave as info for now.

backends/kv/writer.go
gtopper: Will leave as is for now.

backends/kv/writer.go
gtopper: Ok, not linter in Frames apparently.

PR #290, opened 2019-09-24
fix timestamp default value

talIguaz:


column.go
```diff
@@ -356,8 +356,13 @@ func NewSliceColumn(name string, data interface{}) (Column, error) {
 		times := data.([]time.Time)
 		msg.Dtype = pb.DType_TIME
 		msg.Times = make([]int64, len(times))
+		nilTime := time.Time{}
 		for i, t := range times {
-			msg.Times[i] = t.UnixNano()
+			if t == nilTime {
```
gtopper: Might be a good idea to defend against all unsupported values.

talIguaz: Yea i thought about it.. i dont really think it's gonna have other unsupported values. this only happens when we initialize the `times` slice with a certain size so the "zero" values or this shitty thing 0001-01-01...

gtopper: Not sure why it can't have other unsupported values?

PR #291, opened 2019-09-24
[DOC] README: Filter out update-expression refs (not supported) [IG-12272]

Sharon-iguazio:
@dinal please merge.


Sharon-iguazio:
The second commit also fixes the name of the DF variable in the assignment to the `dfs` parameter of the `write` method in the KV write example.

PR #303, opened 2019-11-06
Cache TSDB queriers.

gtopper:
Also reduce log level for per-request logs.

### DIFF
--- a/backends/tsdb/backend.go
+++ b/backends/tsdb/backend.go
@@ -36,6 +36,7 @@ import (
 	"github.com/pkg/errors"
 	"github.com/v3io/frames/v3ioutils"
 	"github.com/v3io/v3io-tsdb/pkg/config"
+	"github.com/v3io/v3io-tsdb/pkg/pquerier"
 	"github.com/v3io/v3io-tsdb/pkg/tsdb"
 	"github.com/v3io/v3io-tsdb/pkg/tsdb/schema"
 	tsdbutils "github.com/v3io/v3io-tsdb/pkg/utils"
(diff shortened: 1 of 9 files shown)


backends/tsdb/backend.go
```diff
@@ -51,8 +57,13 @@ type Backend struct {
 func NewBackend(logger logger.Logger, httpClient *fasthttp.Client, cfg *frames.BackendConfig, framesConfig *frames.Config) (frames.DataBackend, error) {
 
 	frames.InitBackendDefaults(cfg, framesConfig)
+	adapterCacheSize := framesConfig.AdapterCacheSize
+	if adapterCacheSize == 0 {
+		adapterCacheSize = 64
```
talIguaz: why 64?

backends/tsdb/backend.go
```diff
@@ -110,15 +121,44 @@ func (b *Backend) newAdapter(session *frames.Session, password string, token str
...
 
-	adapter, err := b.newAdapter(session, password, token, path)
-	if err != nil {
-		return nil, err
+	h := fnv.New64()
+	_, _ = h.Write(getBytes(session.Url))
+	_, _ = h.Write(getBytes(session.Container))
+	_, _ = h.Write(getBytes(session.Path))
+	_, _ = h.Write(getBytes(session.User))
+	_, _ = h.Write(getBytes(session.Password))
```
talIguaz: shouldn't you use `password, token and path` parameters instead of the session ones?

backends/tsdb/backend.go
gtopper: Because 128 seemed like too many 😄

backends/tsdb/backend.go
```diff
@@ -63,7 +63,7 @@ func NewBackend(logger logger.Logger, httpClient *fasthttp.Client, cfg *frames.B
 	}
 
 	newBackend := Backend{
-		adapters:      lru.New(adapterCacheSize),
+		queriers:      lru.New(adapterCacheSize),
```
talIguaz: change the `adapterCacheSize` name as well

PR #305, opened 2019-11-11
IG-13475 - null values support

talIguaz:


frames.proto
```diff
@@ -58,11 +58,16 @@ message Value {
     }
 }
 
+message RowBitmask {
+    repeated string nullColumns = 1;
```
gtopper: If you don't want to use a bitmask, you don't get to call it a bitmask 😝

clients/py/v3io_frames/pbutils.py
```diff
@@ -226,6 +262,91 @@ def series2col(s, name):
...
+    :return:
+    """
+    bitmask = []
+    for index, row in df.iterrows():
+        current_bitmask = []
+        for col_name in df.columns:
+            if pd.isnull(row[col_name]):
+                df.at[index, col_name] = get_empty_value_by_type(schema[col_name])
+                current_bitmask.append(col_name)
+        bitmask.append({'nullColumns': current_bitmask})
```
gtopper: Just keep a list of lists?

frame.go
```diff
@@ -211,6 +217,17 @@ func (fr *frameImpl) Proto() *pb.Frame {
 	return fr.msg
 }
 
+func (fr *frameImpl) IsNull(index int, colName string) bool {
+	rowBitmask := fr.msg.Bitmask[index]
+	for _, nullColumnName := range rowBitmask.NullColumns {
```
gtopper: If we're going to check for membership and not care about order, shouldn't `NullColumns` be a set?

frame.go
```diff
@@ -398,3 +415,14 @@ func cols2PB(columns []Column) ([]*pb.Column, error) {
 
 	return pbCols, nil
 }
+
+func getKeys(myMap map[string]struct{}) []string {
```
gtopper: This function is not used.

clients/py/v3io_frames/pbutils.py
talIguaz: Can't do. i need an "object" named `nullColumns` otherwise protobuf is unable to create the Frame object

backends/kv/reader.go
```diff
@@ -170,12 +178,15 @@ func (ki *Iterator) Next() bool {
 			}
 
 			var err error
-			err = utils.AppendNil(col)
+			err = utils.AppendNil(byName[name])
```
imakhlin: The implementation of `utils.AppendNil` does not make sense to me
IMHO, you should use complex data types (SQL data type structs) to represent the values, this will allow full and correct support of nullable values

clients/py/v3io_frames/pbutils.py
imakhlin: inconsistent representation of numeric data types. Uninitialised integers will be initialized with type's default value (0) while floats with NaN

backends/kv/reader.go
imakhlin: After F2F discussion we have had with @talIguaz I've convinced that with current implementation of frames the idea of using a map of column indices (map[int]bool) per row is a viable solution.
I also understood that with this solution actual value of the  field that has no value (None/Null) is not important at all, i.e init value in that case could be the default value of corresponding data type.

backends/kv/reader.go
```diff
@@ -170,12 +178,15 @@ func (ki *Iterator) Next() bool {
...
 			if err != nil {
 				ki.err = err
 				return false
 			}
+			currentNullMask.NullColumns[name] = true
```
imakhlin: IMHO it's better to use integer index where possible. In case your data structure has consistent column index - use integer index instead of name for faster access and memory efficiency

backends/kv/reader.go
```diff
@@ -208,8 +219,11 @@ func (ki *Iterator) Next() bool {
 		}
 	}
 
+	if !hasAnyNulls {
+		bitmask = nil
+	}
 	var err error
-	ki.currFrame, err = frames.NewFrame(columns, indices, nil)
+	ki.currFrame, err = frames.NewFrameWithBitmask(columns, indices, nil, bitmask)
```
imakhlin: Rename, how about  NewFrameWithNullValues

clients/py/v3io_frames/pbutils.py
```diff
@@ -127,6 +129,7 @@ def msg2df(frame, frame_factory, columns=None, do_reorder=True):
             else:
                 df = df.reindex(columns=sorted(df.columns))
 
+    df = insert_nulls_based_on_bitmask(df, frame.bitmask)
```
imakhlin: ~~bitmask~~ -> nullmap

clients/py/v3io_frames/pbutils.py
```diff
@@ -226,6 +262,99 @@ def series2col(s, name):
...
+        bitmask.append({'nullColumns': current_bitmask})
+
+    return df, bitmask
+
+
+def get_empty_value_by_type(dtype):
+    if dtype == fpb.INTEGER:
+        return 0
+    elif dtype == fpb.FLOAT:
+        return np.NaN
```
imakhlin: 0.0

frame.go
```diff
@@ -211,6 +222,15 @@ func (fr *frameImpl) Proto() *pb.Frame {
 	return fr.msg
 }
 
+func (fr *frameImpl) IsNull(index int, colName string) bool {
+	if len(fr.msg.Bitmask) == 0 {
+		return false
+	}
+	_, exist := fr.msg.Bitmask[index].NullColumns[colName]
```
imakhlin: use index far all: ~~NullColumns[colName]~~ -> NullColumns[index]

frames.proto
```diff
@@ -58,11 +58,16 @@ message Value {
...
+message RowBitmask {
+    map<string, bool> nullColumns = 1;
+}
+
 message Frame {
     repeated Column columns = 1;
     repeated Column indices = 2;
     map<string, Value> labels = 3;
     string error = 4; // Used in errors when reading over HTTP
+    repeated RowBitmask bitmask = 5;
```
imakhlin: ~~RowBitmask~~ -> NullColumnIndices

test/kv_integration_test.go
```diff
@@ -138,3 +148,169 @@ func (kvSuite *KvTestSuite) TestAll() {
...
+	}
+
+	columns := []frames.Column{
+		floatCol(kvSuite.T(), "n1", len(index)),
+		stringCol(kvSuite.T(), "n2", len(index)),
+		boolCol(kvSuite.T(), "n3", len(index)),
+		timeCol(kvSuite.T(), "n4", len(index)),
+	}
+
+	nullValues := initializeNullColumnsBitmask(len(index))
```
imakhlin: ~~initializeNullColumnsBitmask~~ -> initializeNullColumnsIndex

backends/kv/reader.go
```diff
@@ -111,7 +111,7 @@ func (ki *Iterator) Next() bool {
 
 	rowNum := 0
 	numOfSchemaFiles := 0
-	var bitmask []*pb.RowBitmask
+	var nullColumns []*pb.NullMap
```
imakhlin: minor: NullValuesMap

talIguaz:
Decided to **not** change the null values map from a string "column name" key to an integer "column index" key, since most of the access and search in that map is done by column name.
other then that fixed all other comments including changed naming conventions from `bitmask` to `nullvalues`

PR #309, opened 2019-12-02
IG-13663

talIguaz:
- Support Update mode for KV writes, both frame write and update expression.
- Add `DEBUG` mode for frames integration tests. if you set the env var DEBUG to true, tests will be ran on you local debug frames.
- Add `TEST_BACKENDS` env var to it tests to specify specific backends to test.


backends/kv/writer.go
```diff
@@ -61,14 +63,46 @@ func (kv *Backend) Write(request *frames.WriteRequest) (frames.FrameAppender, er
 		return nil, err
 	}
 
+	var schema v3ioutils.V3ioSchema
+	schema, schemaError := v3ioutils.GetSchema(tablePath, container)
+
+	// Ignore 404 error, since it makes sense there is no schema yet.
+	tableAlreadyExists := true
+	if err != nil {
```
gtopper: Redundant if - `err` cannot be anything but `nil` at this point.

gtopper: I guess you meant to check and use `schemaError`. Better just overwrite `err` I guess, to avoid this kind of confusion.

backends/kv/writer.go
```diff
@@ -61,14 +63,46 @@ func (kv *Backend) Write(request *frames.WriteRequest) (frames.FrameAppender, er
...
+		// If this is the first time we writing to the table, there is nothing to delete.
+		if tableAlreadyExists {
+			err = v3ioutils.DeleteTable(kv.logger, container, tablePath, "", kv.numWorkers, true)
+			if err != nil {
+				return nil, fmt.Errorf("error occured while deleting the table, err: %v", err)
+			}
+			schema = nil
+		}
+	case frames.ErrorIfExists:
+		if schemaError == nil {
```
gtopper: Why not use `tableAlreadyExists` here?

backends/kv/writer.go
```diff
@@ -61,14 +63,46 @@ func (kv *Backend) Write(request *frames.WriteRequest) (frames.FrameAppender, er
...
+		}
+	}
+
+	switch request.SaveMode {
+	case frames.Overwrite:
+		// If this is the first time we writing to the table, there is nothing to delete.
+		if tableAlreadyExists {
+			err = v3ioutils.DeleteTable(kv.logger, container, tablePath, "", kv.numWorkers, true)
+			if err != nil {
+				return nil, fmt.Errorf("error occured while deleting the table, err: %v", err)
```
gtopper: Better say which table in the error message (as you did in case of ErrorIfExists).

backends/kv/writer.go
```diff
@@ -61,14 +63,46 @@ func (kv *Backend) Write(request *frames.WriteRequest) (frames.FrameAppender, er
...
+			}
+			schema = nil
+		}
+	case frames.ErrorIfExists:
+		if schemaError == nil {
+			return nil, fmt.Errorf("table %v already exists, either use a differnet save mode or save to a different table", tablePath)
```
gtopper: Maybe single quote the table name?

types.go
```diff
@@ -38,6 +39,45 @@ var (
...
+		return "overwrite"
+	case Replace:
+		return "replace"
+	default:
+		return ""
+	}
+}
+
+const (
+	ErrorIfExists SaveMode = 0
```
gtopper: Use iota?

types.go
```diff
@@ -38,6 +39,45 @@ var (
...
+func (mode SaveMode) GetNginxModeName() string {
+	switch mode {
+	case ErrorIfExists:
+		return ""
+	case Append:
+		return "CreateNewAttributes"
+	case Overwrite:
+		return ""
+	case Replace:
+		return "CreateOrReplaceAttributes"
```
gtopper: In case of replace, we need to write the nil values so that the attributes get deleted if already present in v3io.

types.go
```diff
@@ -181,3 +222,18 @@ func InitSecretString(s string) SecretString {
...
+	case "", "errorIfExists": // this is the default save mode
+		return ErrorIfExists, nil
+	case "append":
+		return Append, nil
+	case "overwrite":
+		return Overwrite, nil
+	case "replace":
+		return Replace, nil
+	default:
+		return -1, errors.Errorf("no save mode named '%v' ", mode)
```
gtopper: Accidental trailing space in string.

v3ioutils/schema.go
```diff
@@ -130,28 +132,19 @@ func (s *OldV3ioSchema) merge(new *OldV3ioSchema) (bool, error) {
...
 
+	// Do not accept key name change, unless it's the first time we are saving to this table
 	if s.Key != new.Key && new.Key != "" {
-		s.Key = new.Key
-		changed = true
+		if isFirstSchema {
+			s.Key = new.Key
+			changed = true
+		} else {
+			return changed, fmt.Errorf("changing primary key is not allowed, old: %v, new:%v", s.Key, new.Key)
```
gtopper: Missing space after `new:`.

backends/kv/writer.go
```diff
@@ -437,60 +469,110 @@ func (a *Appender) respWaitLoop(timeout time.Duration) {
...
+	return expression.String(), key, sortingVal, nil
+}
+
+func valueToTypedExpressionString(value interface{}, name string) string {
+	switch typedVal := value.(type) {
+	case string:
+		return fmt.Sprintf("%v='%v';", name, typedVal)
+	case time.Time:
+		seconds := typedVal.Unix()
+		nanos := typedVal.UnixNano() % 1e9
```
gtopper: Use `nanos := typedVal.Nanosecond()`❓

backends/kv/writer.go
```diff
@@ -437,60 +469,110 @@ func (a *Appender) respWaitLoop(timeout time.Duration) {
...
+		expression.WriteString(valueToTypedExpressionString(sortingVal, sortingKeyName))
+	}
+
+	return expression.String(), key, sortingVal, nil
+}
+
+func valueToTypedExpressionString(value interface{}, name string) string {
+	switch typedVal := value.(type) {
+	case string:
+		return fmt.Sprintf("%v='%v';", name, typedVal)
```
gtopper: What if the string is `'`?

clients/py/v3io_frames/client.py
```diff
@@ -139,6 +140,10 @@ def write(self, backend, table, dfs, expression='', condition='',
         index_cols : []str
             List of column names to be used as the index columns for the write
             operation; by default, the DataFrame's index columns are used
+        save_mode : str
+            Save mode.
+            Optional values: errorIfExists (default), append,
+             overwrite, replace
```
gtopper: I thought we had 5 modes? 😄

types.go
```diff
@@ -38,6 +39,50 @@ var (
...
+func (mode SaveMode) GetNginxModeName() string {
+	switch mode {
+	case ErrorIfTableExists:
+		return ""
+	case OverwriteTable:
+		return ""
+	case UpdateItem:
+		return "CreateOrReplaceAttributes"
+	case OverwriteItem:
+		return "OverWriteAttributes"
```
gtopper: sic?

talIguaz: yap

backends/kv/writer.go
talIguaz: guess there is nothing we can do

backends/kv/reader.go
```diff
@@ -169,7 +163,13 @@ func (ki *Iterator) Next() bool {
 				colName = ki.schema.Key
 			}
 
-			col := byName[colName]
+			col, ok := byName[colName]
+			if !ok {
+				ki.err = fmt.Errorf("column '%v' for item with key: '%v' does not exist in the schema file. "+
+					"your data structure was probably changed, try re-inferring the schema for the table",
```
gtopper: `your` -> `Your`

backends/kv/writer.go
```diff
@@ -92,28 +124,23 @@ func (a *Appender) Add(frame frames.Frame) error {
...
-		return err
-	}
-
 	if a.request.Expression != "" {
 		return a.update(frame)
 	}
 
 	columns := make(map[string]frames.Column)
-	indexName := ""
+	indexName, sortingKeyName := "", ""
```
gtopper: `var indexName, sortingKeyName string`❔

backends/kv/writer.go
gtopper: Not currently anyway 🕵🏻‍♂️

PR #310, opened 2019-12-04
IG-13749 Range scan

dinal:


test/kv_integration_test.go
```diff
@@ -149,6 +149,77 @@ func (kvSuite *KvTestSuite) TestAll() {
 
 }
 
+func (kvSuite *KvTestSuite) TestRangeScan() {
```
talIguaz: * validate the schema was created successfully.
* more tests + negative
* negative tests that writes multiple frames to the same table

backends/kv/reader.go
```diff
@@ -70,11 +70,12 @@ func (kv *Backend) Read(request *frames.ReadRequest) (frames.FrameIterator, erro
 		true,
 		len(partitions)*numberOfWorkers)
 
-	input := v3io.GetItemsInput{Filter: request.Proto.Filter, AttributeNames: columns}
+	input := v3io.GetItemsInput{Filter: request.Proto.Filter, AttributeNames: columns, SortKeyRangeStart: request.Proto.SortKeyRangeStart, SortKeyRangeEnd: request.Proto.SortKeyRangeEnd}
```
talIguaz: We need to also add the sorting key (if exists) to the returned DF index.

v3ioutils/schema.go
```diff
@@ -154,6 +155,11 @@ func (s *OldV3ioSchema) merge(new *OldV3ioSchema) (bool, error) {
 		changed = true
 	}
 
+	if s.SortingKey != new.SortingKey && new.SortingKey != "" {
```
talIguaz: What if i had a table with sorting key `colA` then writing a new DF without a sorting key ? looks like you override the schema with an empty sortingKey which will corrupt the data. you should return an error in that case.
I did some changes for that in my PR probably should use that 😄

backends/kv/writer.go
```diff
@@ -196,6 +213,16 @@ func (a *Appender) Add(frame frames.Frame) error {
 	return nil
 }
 
+func (a *Appender) formatKeyName(indicesSize int, key interface{}, sortingVal interface{}) string {
+	var format string
+	if indicesSize > 1 {
```
talIguaz: you can just check if `sortingVal !=nil`

backends/kv/writer.go
```diff
@@ -181,7 +185,20 @@ func (a *Appender) Add(frame frames.Frame) error {
...
 		}
 
-		input := v3io.PutItemInput{Path: a.tablePath + fmt.Sprintf("%v", key), Attributes: row, Condition: condition}
+		key := indexVal(r)
+		// Add key column as an attribute
+		row[indexName] = key
+
+		var sortingVal interface{}
+		if len(indices) > 1 {
+			sortingFunc, err := a.funcFromCol(indices[1])
```
talIguaz: save this function outside the loop so that it will only be called once.

backends/kv/writer.go
```diff
@@ -181,7 +185,20 @@ func (a *Appender) Add(frame frames.Frame) error {
...
+			if err != nil {
+				return err
+			}
+			sortingVal = sortingFunc(r)
+			row[indices[1].Name()] = sortingVal
```
talIguaz: can use `row[schema.SortingKey] = sortingVal`

backends/kv/writer.go
```diff
@@ -225,7 +252,16 @@ func (a *Appender) update(frame frames.Frame) error {
 		}
 
 		key := indexVal(r)
-		input := v3io.UpdateItemInput{Path: a.tablePath + fmt.Sprintf("%v", key), Expression: expr, Condition: cond}
+		var sortingVal interface{}
+		if len(frame.Indices()) > 1 {
+			sortingFunc, err := a.funcFromCol(frame.Indices()[1])
```
talIguaz: same thing as before, get the `sortingFunc` declaration out of the loop

backends/kv/writer.go
dinal: to use schema i need to do  "newSchema.(*v3ioutils.OldV3ioSchema).SortingKey", i prefer the current way

backends/kv/reader.go
```diff
@@ -203,20 +205,20 @@ func (ki *Iterator) Next() bool {
...
+		if len(columns) > 1 || columns[0].Name() != ki.schema.Key {
+			shouldDelete := ki.handleIndices(ki.schema.Key, byName, ki.shouldDuplicateIndex, &indices)
+			if shouldDelete {
 				columns = utils.RemoveColumn(ki.schema.Key, columns)
 			}
+			delete(byName, ki.schema.Key)
+		}
+		if  ki.schema.SortingKey != "" && (len(columns) > 1 || columns[0].Name() != ki.schema.SortingKey) {
+			shouldDelete := ki.handleIndices(ki.schema.SortingKey, byName, ki.shouldDuplicateSorting, &indices)
+			if shouldDelete {
```
talIguaz: Why not do this in the `handleIndices` as well?

backends/kv/writer.go
talIguaz: ok.. you still can do this once in the beginning of the function and avoid redundant function calls on every iteration.. but 🤷‍♂

dinal: it needs to be inside the loop as the value is calculated with the loop index (r)

PR #313, opened 2019-12-15
Ig 14072

dinal:


test/kv_integration_test.go
```diff
@@ -524,3 +524,61 @@ func (kvSuite *KvTestSuite) TestRequestSpecificColumnsWithKey() {
...
+	}
+
+	it, err := kvSuite.client.Read(rreq)
+	if err != nil {
+		kvSuite.T().Fatal(err)
+	}
+
+	for it.Next() {
+		frame := it.At()
+		if frame.Len() != 0 {
```
talIguaz: better use `kvSuite.Require.Equals(<expected>, <actual> , <error message>)`

test/kv_integration_test.go
```diff
@@ -524,3 +524,61 @@ func (kvSuite *KvTestSuite) TestRequestSpecificColumnsWithKey() {
...
+			kvSuite.T().Fatalf("wrong length: %d != %d", frame.Len(), 0)
+		}
+	}
+	//make sure schema is not deleted
+	schemaInput := &v3io.GetObjectInput{Path: table + "/.#schema"}
+	_, err = kvSuite.v3ioContainer.GetObjectSync(schemaInput)
+	if err != nil {
```
talIguaz: Also here, `kvSuite.NoError(err, "schema is not found " + err.Error())`

v3ioutils/container.go
```diff
@@ -198,19 +201,22 @@ func getItemsWorker(container v3io.Container, input *v3io.GetItemsInput, fileNam
...
 			if !ok {
 				terminationChan <- nil
 				return
 			}
-			input := &v3io.DeleteObjectInput{Path: tablePath + "/" + fileName}
-			err := container.DeleteObjectSync(input)
-			if err != nil {
-				terminationChan <- err
-				return
+			if deleteSchema || (!deleteSchema && fileName != ".#schema") {
```
talIguaz: redundant check for the second `!deleteSchema`. you can just write:
`if deleteSchema || fileName != ".#schema"`

talIguaz: Or maybe better, to filter out the schema file in `getItemsWorker` and here to delete everything that is returned without any conditions (might be a bit faster 🤷‍♂ ).

PR #314, opened 2019-12-16
IG 14140

dinal:


backends/kv/reader.go
```diff
@@ -148,6 +149,19 @@ func (ki *Iterator) Next() bool {
 		byName[field.Name] = col
 	}
 
+	if specificColumnsRequested {
+		for _, attr := range systemAttrs {
+			if containsString(ki.request.Proto.Columns, attr) {
```
talIguaz: why not check starts with double underscore?

backends/kv/reader.go
```diff
@@ -148,6 +149,19 @@ func (ki *Iterator) Next() bool {
...
+				sysCol, err := frames.NewSliceColumn(attr, make([]int64, 0))
```
talIguaz: Are we sure all of the system attribute are/will be of type int64?

backends/kv/reader.go
```diff
@@ -148,6 +149,19 @@ func (ki *Iterator) Next() bool {
 		byName[field.Name] = col
 	}
 
+	if specificColumnsRequested {
```
talIguaz: Better add a condition to see if the `len(columnNamesToReturn) != len(columns)` to avoid iterating the systemAttribute for no reason

backends/kv/reader.go
dinal: yes, they are all int64 except __name (verified with ortal)

backends/kv/reader.go
```diff
@@ -22,7 +22,6 @@ package kv
 
 import (
 	"fmt"
-
```
talIguaz: add the new line

backends/kv/reader.go
```diff
@@ -157,10 +185,11 @@ func (ki *Iterator) Next() bool {
 		for name, field := range row {
 			colName := name
 			if colName == indexColKey { // convert `__name` attribute name to the key column
-				if hasKeyColumnAttribute {
+				if hasKeyColumnAttribute && !indexKeyRequested {
```
talIguaz: move `&& !indexKeyRequested` to the outer if for clarity.

test/kv_integration_test.go
```diff
@@ -646,3 +646,42 @@ func (kvSuite *KvTestSuite) TestDeleteWithFilter() {
...
+
+	err = appender.Add(frame)
+	kvSuite.NoError(err)
+
+	err = appender.WaitForComplete(3 * time.Second)
+	kvSuite.NoError(err)
+
+	time.Sleep(3 * time.Second) // Let DB sync
+
+	requestedColumns := []string{"n1", "__name", "__gid", "__mode", "__mtime_nsecs", "__mtime_secs", "__size", "__uid", "__ctime_nsecs", "__ctime_secs"}
```
talIguaz: maybe add the index to the columns as well?

backends/kv/reader.go
```diff
@@ -185,10 +185,12 @@ func (ki *Iterator) Next() bool {
 		for name, field := range row {
 			colName := name
 			if colName == indexColKey { // convert `__name` attribute name to the key column
-				if hasKeyColumnAttribute && !indexKeyRequested {
-					continue
-				} else if !hasKeyColumnAttribute && !indexKeyRequested {
-					colName = ki.schema.Key
+				if !indexKeyRequested {
```
talIguaz: add this to the outer if  `if colName == indexColKey  && !indexKeyRequested` 😅

PR #318, opened 2019-12-18
Add profiling endpoint

talIguaz:


cmd/framesd/framesd.go
```diff
@@ -97,6 +98,26 @@ func main() {
...
+
+			if hasLog {
+				logger.Info("creating profiling endpoint at :8082. To start profiling run one of the following commands:\n" +
+					" -> profiling: curl -o cpu.pprof localhost:8082/debug/pprof/profile?seconds=30\n" +
+					" -> trace: curl -o trace.pprof localhost:8082/debug/pprof/trace?seconds=30")
+			}
+
+			err := http.ListenAndServe(":8082", nil)
+			if err != nil && hasLog {
+				logger.Warn("failed to create profiling endpoint")
```
gtopper: Do include the error in the log, so we can see what happened (e.g. the port was taken).

PR #321, opened 2019-12-22
IG-12619: Fix infer schema

gtopper:


backends/kv/inferschema.go
```diff
@@ -35,57 +39,140 @@ func (b *Backend) inferSchema(request *frames.ExecRequest) error {
...
 	for rowNum := 0; rowNum < maxrec && iter.Next(); rowNum++ {
 		row := iter.GetFields()
 		rowSet = append(rowSet, row)
-		index, ok := row["__name"]
+		key, ok := row["__name"]
 		if !ok {
 			return fmt.Errorf("key (__name) was not found in row")
 		}
-		indicies = append(indicies, index.(string))
+		keys = append(keys, key.(string))
```
dinal: you're not doing anything with 'keys'

backends/kv/inferschema.go
```diff
@@ -35,57 +39,140 @@ func (b *Backend) inferSchema(request *frames.ExecRequest) error {
...
+		} else if len(possibleFullKeys) == 1 {
+			keyField = possibleFullKeys[0]
+			sortingKeyField = ""
+		} else {
+			var reason string
+			if len(possibleFullKeys) == 0 {
+				reason = "no column matched __name attribute"
+			} else {
+				sort.Strings(possibleFullKeys)
+				reason = fmt.Sprintf("%d columns (%s) matched __name attribute", len(possibleFullKeys), strings.Join(possibleFullKeys, ", "))
```
dinal: not sure user is familiar with __name, better rephrase..

backends/kv/inferschema.go
```diff
@@ -35,57 +39,140 @@ func (b *Backend) inferSchema(request *frames.ExecRequest) error {
...
+					columnCanBeSortingKey[attrName] = true
+				}
+				columnCanBeSortingKey[attrName] = columnCanBeSortingKey[attrName] && attrValue == sortingKeyValue
+			}
+		}
+	}
+
+	var primaryKeyField string
+	var sortingKeyField string
+	if keyField == "" {
```
dinal: if keyField != "" then no need to do infer, you can check in the start and skip the entire logic

gtopper: Still need to infer to get the type of each column.

PR #333, opened 2019-12-25
IG-13619 handle non existing column

dinal:


test/kv_integration_test.go
```diff
@@ -684,4 +684,37 @@ func (kvSuite *KvTestSuite) TestRequestSystemAttrs() {
...
+	kvSuite.T().Log("read")
+	rreq := &pb.ReadRequest{
+		Backend: kvSuite.backendName,
+		Table:   table,
+		Columns: []string{"n1", "kk"},
+	}
+
+	it, _ := kvSuite.client.Read(rreq)
+	it.Next()
+	kvSuite.NotEmptyf(it.Err(), "error was expected when reading a non existing column")
```
talIguaz: better use `kvSuite.Error(it.Err(),  "error was expected when reading a non existing column" )`

PR #343, opened 2020-01-01
IG-14367 Api changes

dinal:


backends/stream/backend.go
```diff
@@ -62,25 +61,20 @@ func (b *Backend) Create(request *frames.CreateRequest) error {
...
-		if !isInt || shards < 1 {
-			return errors.Errorf("Shards attribute must be a positive integer (got %v)", shardsVar)
+	if request.Proto.Shards != 0 {
+		shards = request.Proto.Shards
+		if shards < 1 {
+			return errors.Errorf("Shards attribute must be a positive integer (got %v)", shards)
 		}
 	}
 
 	retention := int64(24)
```
gtopper: `var retention int64 = 24`

gtopper: Oh well, it was there before this PR 😄

clients/py/tests/conftest.py
```diff
@@ -43,6 +43,7 @@
 
 extra_backends = [
     {'type': 'kv'},
+    {'type': 'mysql'},
```
meirleopold: did you mean nosql?

f.        [#5] Add a "nosql" alias to the "kv" backend type to match the terminology used in the docs. (I suggest adding an alias rather than renaming to avoid breaking backwards compatibility.)

dinal: oh no lol
i'll fix it

PR #390, opened 2020-02-23
[DOC] get_raw read param - for internal use only [IG-14773]

Sharon-iguazio:
Per R&D's request, remove the detailed `read` method `get_raw` parameter documentation and mark it for internal use only. @dinal please review and merge.


clients/py/v3io_frames/client.py
```diff
@@ -235,6 +235,8 @@ def read(self, backend='', table='', query='', columns=None, filter='',
             True - return a DataFrames iterator;
             False (default except for get_raw=True) - return a single DataFrame
         get_raw (Optional) : bool
+            True to return the data in raw format instead as pandas DataFrames
```
dinal: no need for this line, it's mentioned 2 lines down

Sharon-iguazio: @dinal This line was actually intended to replace the existing detailed description, but I see now that I forgot to delete that description. => Done now in commit af138f7.

PR #391, opened 2020-02-24
Added linting. Changed code to conform to linter rules.

gtopper:


backends/kv/reader.go
```diff
@@ -66,6 +66,10 @@ func (kv *Backend) Read(request *frames.ReadRequest) (frames.FrameIterator, erro
 		request.Proto.Table,
 		true)
 
+	if err != nil {
+		return nil, err
```
gtopper: Semantic change❗️ Error check was missing.

PR #393, opened 2020-02-24
create num workers based on num vns

dinal:


go.sum
```diff
@@ -10,6 +10,9 @@ github.com/cpuguy83/go-md2man v1.0.10/go.mod h1:SmD6nW6nTyfqj6ABTjUi3V3JVMnlJmwc
 github.com/davecgh/go-spew v1.1.0/go.mod h1:J7Y8YcW2NihsgmVo/mv3lAwl/skON4iLHjSsI+c5H38=
 github.com/davecgh/go-spew v1.1.1 h1:vj9j/u1bqnvCEfJOwUhtlOARqs3+rkHYY13jYWTU97c=
 github.com/davecgh/go-spew v1.1.1/go.mod h1:J7Y8YcW2NihsgmVo/mv3lAwl/skON4iLHjSsI+c5H38=
+github.com/dinal/v3io-go v0.0.5-0.20200304161325-3f0c978b8680 h1:UQ9aRIFGgbYiwKQWBoNYBBeqoXjhHBFu7JeWm+l/YeE=
+github.com/dinal/v3io-go v0.0.5-0.20200304161325-3f0c978b8680/go.mod h1:D0W1tjsVgcp4xk3ZI2fjKTKaOpYJLewN1BPN0x2osO4=
+github.com/dinal/v3io-go v0.1.5-0.20200301152134-6880d30985de/go.mod h1:D0W1tjsVgcp4xk3ZI2fjKTKaOpYJLewN1BPN0x2osO4=
```
gtopper: `go mod tidy`

backends/kv/backend.go
```diff
@@ -123,5 +122,15 @@ func (b *Backend) newConnection(session *frames.Session, password string, token
...
 		b.logger)
 
+	if err == nil && b.numWorkers == 0 {
+		resp, err := container.GetClusterMDSync(&v3io.GetClusterMDInput{})
+		if err != nil {
+			return nil, "", fmt.Errorf("could not detrmine num vns in cluster")
+		}
+		getClusterMDOutput := resp.Output.(*v3io.GetClusterMDOutput)
+		b.numWorkers = getClusterMDOutput.NumberOfVNs
+		b.logger.Debug("going to use ", b.numWorkers, "workers")
```
gtopper: Wrong usage of `logger.Debug`.  You probably want to use:
```
b.logger.DebugWith("setting number of workers", "workers", b.numWorkers)
```
or 
```
b.logger.Debug("setting number of workers to %d", b.numWorkers)
```

backends/kv/backend.go
```diff
@@ -123,5 +122,15 @@ func (b *Backend) newConnection(session *frames.Session, password string, token
 		token,
 		b.logger)
 
+	if err == nil && b.numWorkers == 0 {
```
gtopper: Wouldn't it make more sense to do this earlier, when we create the backend or the v3io session?

gtopper: I suppose the downside(?) would be that it would require cluster connectiity to start frames.

backends/tsdb/backend.go
```diff
@@ -111,6 +111,15 @@ func (b *Backend) newAdapter(session *frames.Session, password string, token str
 		return nil, err
 	}
 
+	if b.backendConfig.Workers == 0 {
+		resp, err := container.GetClusterMDSync(&v3io.GetClusterMDInput{})
```
gtopper: This will check the number of VNs on every write or delete.

backends/kv/backend.go
dinal: i need user pass or token to open a connection which i don't have on startup, need to wait for a req and use it from there

backends/tsdb/backend.go
dinal: why? b.backendConfig.Workers won't be 0 on the second time here

PR #400, opened 2020-02-25
IG-14876: KV table is created by Frame client without sharding key

imakhlin:
stop writing dataframe if one of the index columns has empty value


backends/kv/writer.go
```diff
@@ -211,6 +221,13 @@ func (a *Appender) Add(frame frames.Frame) error {
 			return err
 		}
 
+		if indexName != "" && keyVal == "" {
```
talIguaz: i think `indexName` can't be empty

test/kv_save_mode_integration_test.go
```diff
@@ -205,6 +206,29 @@ func (kvSuite *KvTestSuite) TestSaveModeUpdateItemNewRow() {
 	kvSuite.NoError(iter.Err(), "error reading from kv")
 }
 
+func (kvSuite *KvTestSuite) TestSaveModeUpdateItemNoKey() {
```
talIguaz: This test is not relevant to save modes test..

test/kv_integration_test.go
```diff
@@ -66,6 +66,32 @@ func (kvSuite *KvTestSuite) SetupSuite() {
 	}
 }
 
+func (kvSuite *KvTestSuite) generateRandomSampleFrameWithEmptyStringIndex(size int, indexName string, columnNames []string) frames.Frame {
```
talIguaz: how about tests for sorting key has an empty value

test/kv_integration_test.go
```diff
@@ -66,25 +67,37 @@ func (kvSuite *KvTestSuite) SetupSuite() {
...
+func (kvSuite *KvTestSuite) generateRandomSampleFrameWithEmptyIndices(size int, indexNames []string, emptyIndices map[string]bool, columnNames []string) frames.Frame {
+	kvSuite.Require().True(len(indexNames) <= 2, "KV API supports no more than two indices")
+	var indexColumns []frames.Column
 
-	index := make([]string, size)
+	emptyIndexValues := make([]string, size)
+	uniqueIndexValues := make([]string, size)
 	for i := 0; i < size; i++ {
-		index[i] = ""
+		emptyIndexValues[i] = fmt.Sprintf("%d", i)
```
talIguaz: isn't this the `uniqueIndexValues`?

talIguaz: also this means that the errors you checked for are not right.. maybe add a more specific  error check then `"should not be empty"`

PR #405, opened 2020-03-04
IG-14809 - return an empty dataframe for empty dataset

imakhlin:
- [x] return an empty data-frame (without columns metadata) in case of empty dataset for both GRPC and HTTP


clients/py/v3io_frames/client.py
```diff
@@ -409,7 +409,7 @@ def execute(self, backend, table, command='', args=None):
             On request error or backend error
         """
         self._validate_request(backend, table, ExecuteError)
-        return self._execute(self._alias_backends(backend), table, command, args)
+        return self._execute(self._alias_backends(backend), table, command, args, expression=None)
```
talIguaz: why?

imakhlin: because of error:
`E           TypeError: _execute() missing 1 required positional argument: 'expression'`

PR #406, opened 2020-03-04
IG-12645: Fix grafana response column order.

gtopper:
Before this change, the response column order is nondeterministic.

After this change, response columns are in the requested order, or sorted lexicographically in case of a wildcard request.


imakhlin:
The fix looks good.
Could you add an integration test to make sure that we cover the field order scenario with test

PR #411, opened 2020-03-08
IG-15061 (unverified): Avoid goroutine leak on error.

gtopper:
It's bad practice to return both a value and an error, especially when the value requires closing.


backends/tsdb/writer.go
```diff
@@ -53,7 +53,8 @@ func (b *Backend) Write(request *frames.WriteRequest) (frames.FrameAppender, err
 	if request.ImmidiateData != nil {
 		err := newTsdbAppender.Add(request.ImmidiateData)
 		if err != nil {
-			return &newTsdbAppender, err
+			newTsdbAppender.Close()
```
gtopper: This is the line that fixes the leak.

PR #425, opened 2020-03-19
IG-13538 - support writing partitioned KV tables

talIguaz:


backends/kv/writer.go
```diff
@@ -238,7 +268,7 @@ func (a *Appender) Add(frame frames.Frame) error {
 			}
 		}
 
-		input := v3io.UpdateItemInput{Path: a.tablePath + a.formatKeyName(keyVal, sortingKeyVal),
+		input := v3io.UpdateItemInput{Path: a.tablePath + subPathString + a.formatKeyName(keyVal, sortingKeyVal),
```
gtopper: Use `Sprintf`?

PR #438, opened 2020-03-31
ignore 404 and 409 errors in delete

dinal:


v3ioutils/container.go
```diff
@@ -145,10 +145,12 @@ func DeleteTable(logger logger.Logger, container v3io.Container, path, filter st
 			}
 		case err := <-deleteTerminationChan:
 			if err != nil {
-				for i := 0; i < 2*workers; i++ {
-					onErrorTerminationChannel <- struct{}{}
+				if errorWithStatusCode, ok := err.(v3ioerrors.ErrorWithStatusCode); !ok || !ignoreMissing || (errorWithStatusCode.StatusCode() != http.StatusNotFound && errorWithStatusCode.StatusCode() != http.StatusConflict) {
```
talIguaz, line 148: 1. use `utils.IsNotExistsOrConflictError(err)` instead of explicitly checking error codes.
2. In this place you will ignore those errors but you stop deleting remaining object in the "bad" partition. i think we better ignore the errors further down at `deleteEntirePartition` and `utils.DeleteTable`

v3ioutils/container.go
```diff
@@ -206,8 +206,10 @@ func deleteObjectWorker(tablePath string, container v3io.Container, fileNameChan
 			input := &v3io.DeleteObjectInput{Path: tablePath + "/" + fileName}
 			err := container.DeleteObjectSync(input)
 			if err != nil {
-				terminationChan <- err
-				return
+				if utils.IsNotExistsOrConflictError(err) {
```
talIguaz, line 209: `if !utils.IsNotExistsOrConflictError(err)`

PR #443, opened 2020-04-07
IG-13538 - support writing partitioned KV tables

talIguaz:


talIguaz:
was reviewed in - https://github.com/v3io/frames/pull/425

PR #446, opened 2020-04-13
IG-15440 fail writing to an existing folder without schema

dinal:


backends/kv/writer.go
```diff
@@ -117,6 +121,12 @@ func (kv *Backend) Write(request *frames.WriteRequest) (frames.FrameAppender, er
 	return &appender, nil
 }
 
+func checkPathExists(tablePath string, container v3io.Container) bool {
+	input := &v3io.CheckPathExistsInput{Path: tablePath}
+	err := container.CheckPathExistsSync(input)
+	return err == nil
```
gtopper, line 127: It could be an error other than 404 status...

backends/kv/writer.go
```diff
@@ -89,6 +89,10 @@ func (kv *Backend) Write(request *frames.WriteRequest) (frames.FrameAppender, er
 		case frames.ErrorIfTableExists:
 			return nil, fmt.Errorf("table '%v' already exists; either use a differnet save mode or save to a different table", tablePath)
 		}
+	} else {
+		if checkPathExists(tablePath, container) {
+			return nil, fmt.Errorf("folder '%v' already exists; you can't write to a non empty folder without a schema file", tablePath)
```
gtopper, line 94: How do we know it's non empty?

go.mod
```diff
@@ -21,4 +21,7 @@ require (
 	google.golang.org/grpc v1.17.0
 )
 
-replace github.com/xwb1989/sqlparser => github.com/v3io/sqlparser v0.0.0-20190306105200-4d7273501871
+replace (
+	github.com/v3io/v3io-go => github.com/dinal/v3io-go v0.0.5-0.20200413110513-3adb776f958c
```
gtopper, line 25: Remider to remove this, update v3io-go version, and go mod tidy after https://github.com/v3io/v3io-go/pull/62 is merged.

backends/kv/writer.go
dinal: we don't, wrong err msg :)

backends/kv/writer.go
```diff
@@ -89,6 +89,14 @@ func (kv *Backend) Write(request *frames.WriteRequest) (frames.FrameAppender, er
 		case frames.ErrorIfTableExists:
 			return nil, fmt.Errorf("table '%v' already exists; either use a differnet save mode or save to a different table", tablePath)
 		}
+	} else {
+		exists, err := checkPathExists(tablePath, container)
+		if err != nil {
+			return nil, err
+		}
+		if exists {
+			return nil, fmt.Errorf("folder '%v' already exists; you can't write to an exisitng folder without a schema file", tablePath)
```
gtopper, line 98: exisitng -> existing

gtopper: To avoid confusion:
without a schema file -> unless it already contains a schema file

PR #448, opened 2020-04-20
IG 15354 - frames history logs

talIguaz:


backends/utils/history_server.go
```diff
@@ -0,0 +1,363 @@
...
+			}
+		}
+	}()
+}
+
+func (m *HistoryServer) createDefaultV3ioClient() {
+
+	session := &frames.Session{}
+
+	token := os.Getenv("V3IO_ACCESS_KEY") // "0938b76f-79ff-41f6-981d-864ea3291cca"
```
gtopper, line 98: Is this your env's access key? 🙈

backends/utils/history_server.go
```diff
@@ -0,0 +1,363 @@
...
+	if token == "" {
+		m.logger.Warn("can not create v3io.client. could not find `V3IO_ACCESS_KEY` environment variable")
```
gtopper, line 100: So if there's no token we'll just warn about it and keep going?

gtopper: Ok, I see the isActive thing now... Why not report the error though?

backends/utils/history_server.go
```diff
@@ -0,0 +1,363 @@
...
+			ActionDuration: duration,
+			ActionType:     queryType,
+			AdditionalData: readRequest.ToMap(),
+			Container:      readRequest.Proto.Session.Container}
+
+		m.requests <- entry
+	}()
+}
+
+func (m *HistoryServer) GetLogs(request *frames.HistoryRequest) (frames.Frame, error) {
```
gtopper, line 158: This method doesn't actually seem to have anything to do with `HistoryServer`.

backends/utils/history_server.go
```diff
@@ -0,0 +1,363 @@
...
+
+	for _, col := range customColumnsByName {
+		allColumns = append(allColumns, col)
+	}
+
+	return frames.NewFrame(allColumns, []frames.Column{indexColumn}, nil)
+}
+
+func (m *HistoryServer) writeMonitoringBatch(logs []HistoryEntry) {
+	d, err := json.Marshal(logs)
```
gtopper, line 309: Wouldn't it make more sense to have one entry per line?

backends/utils/history_server.go
```diff
@@ -0,0 +1,363 @@
...
+	if err != nil {
+		m.logger.Warn("Failed to marshal logs to json, err: %v. logs: %v", err, logs)
+		return
+	}
+
+	d = append(d, []byte("\n")...)
+	input := &v3io.PutObjectInput{Path: m.logFilePath, Body: d, Append: true}
+	err = m.container.PutObjectSync(input)
+	if err != nil {
+		m.logger.Warn("Failed to append Frames History logs to file, err: %v. logs: %v", err, logs)
```
gtopper, line 319: I'm not sure discarding the logs and moving on is the right thing to do. It's not like it's the logs' fault.

backends/utils/history_server.go
talIguaz: hmm yea thought about this.. maybe yes. this will potentially reduce the number of `unmarshals` that we do. you think it's better just a single entry per line?

backends/utils/history_server.go
talIguaz: well i tried to do the entire "error handling" of the ingestion of the logs kinda seamless. because:
1. its async and not part of a user action where you can return an error.
2. it shouldn't effect any normal use of the service. hence i thought for now just logging it should be good enough.

do you have other suggestions?

backends/utils/history_server.go
gtopper: Yup

backends/utils/history_server.go
```diff
@@ -0,0 +1,363 @@
...
+	session := &frames.Session{}
+
+	token := os.Getenv("V3IO_ACCESS_KEY") // "0938b76f-79ff-41f6-981d-864ea3291cca"
+	if token == "" {
+		m.logger.Warn("can not create v3io.client. could not find `V3IO_ACCESS_KEY` environment variable")
+		return
+	}
+
+	var err error
+	m.container, err = m.createV3ioClient(session, "", token)
```
gtopper, line 105: Avoid creating per request to avoid go routine leak.

talIguaz: i am creating the fasthttp client and v3io-go context only once

backends/utils/history_server.go
```diff
@@ -0,0 +1,526 @@
...
+	frame, err := frames.NewFrameWithNullValues(columns, []frames.Column{indexColumn}, nil, nullMask)
+	if err != nil {
+		return nil, err
+	}
+
+	return frame, nil
+}
+
+func (m *HistoryServer) writeMonitoringBatch(logs []HistoryEntry) {
+	var data []byte
```
gtopper, line 425: Consider reusing this one across calls.

backends/utils/history_server.go
```diff
@@ -0,0 +1,526 @@
...
+		data = append(data, []byte("\n")...)
+	}
+
+	logFilePath := m.getCurrentLogFileName()
+	input := &v3io.PutObjectInput{Path: logFilePath, Body: data, Append: true}
+	err := m.container.PutObjectSync(input)
+
+	if err != nil {
+		// retry on 5xx errors
+		if errWithStatusCode, ok := err.(v3ioerrors.ErrorWithStatusCode); ok && errWithStatusCode.StatusCode() > 500 {
```
gtopper, line 444: `>= 500`

backends/utils/history_server.go
```diff
@@ -0,0 +1,526 @@
...
+	if m.config.HistoryFileNum != 0 {
+		m.HistoryFileNum = m.config.HistoryFileNum
+	}
+}
+
+func (m *HistoryServer) getCurrentLogFileName() string {
+	// Round current time to the nearest log time-bucket
+	currentTimeInSeconds := time.Now().Unix()
+	currentTimeBucket := m.HistoryFileDurationSecondSpans * (currentTimeInSeconds / m.HistoryFileDurationSecondSpans)
+	return fmt.Sprintf("%v/%v.json", m.LogsFolderPath, currentTimeBucket)
```
gtopper, line 120: Might be more user friendly to use an ISO timestamp than unix time for the file name.

backends/utils/history_server.go
```diff
@@ -0,0 +1,526 @@
...
+
+	for _, log := range logs {
+		d, err := json.Marshal(log)
+		if err != nil {
+			m.logger.Warn("Failed to marshal logs to json, err: %v. logs: %v", err, logs)
+			return
+		}
+
+		data = append(data, d...)
+		data = append(data, []byte("\n")...)
```
gtopper, line 435: Just `data = append(data, '\n')`?

backends/utils/history_server.go
```diff
@@ -0,0 +1,526 @@
...
+	return frame, nil
+}
+
+func (m *HistoryServer) writeMonitoringBatch(logs []HistoryEntry) {
+	var data []byte
+
+	for _, log := range logs {
+		d, err := json.Marshal(log)
+		if err != nil {
+			m.logger.Warn("Failed to marshal logs to json, err: %v. logs: %v", err, logs)
```
gtopper, line 430: Don't log all the logs... Just the one that failed the marshaller.

backends/utils/history_server.go
```diff
@@ -0,0 +1,526 @@
...
+
+				customColumnsByName = make(map[string][]string)
+				nullColumns = []*pb.NullValuesMap{}
+				hasNullsInCurrentDF = false
+			}
+
+			var entry HistoryEntry
+
+			if err := json.Unmarshal(lineIterator.At(), &entry); err != nil {
+				return fmt.Errorf("error reading logs. json marshal error:%v", err)
```
gtopper, line 303: Better: `failed to parse json: %v`. Even better, include the log that failed the parser.

backends/utils/history_server.go
gtopper: Also better warn about this 💩

backends/utils/history_server.go
```diff
@@ -0,0 +1,526 @@
...
+
+	if err != nil {
+		// retry on 5xx errors
+		if errWithStatusCode, ok := err.(v3ioerrors.ErrorWithStatusCode); ok && errWithStatusCode.StatusCode() > 500 {
+			input := &v3io.PutObjectInput{Path: logFilePath, Body: data, Append: true}
+			err = m.container.PutObjectSync(input)
+		}
+
+		if err != nil {
+			m.logger.Error("Failed to append Frames History logs to file, err: %v. logs: %v", err, logs)
```
gtopper, line 450: Here too, only log the log we failed on, not all of them. Not sure why history is capitalized 🤓

talIguaz: it writes the whole batch of logs, so the whole request failed

api/api.go
```diff
@@ -222,15 +250,29 @@ func (api *API) Exec(request *frames.ExecRequest) (frames.Frame, error) {
...
 	return frame, nil
 }
 
+func (api *API) History(request *frames.HistoryRequest, out chan frames.Frame) error {
+	err := api.historyServer.GetLogs(request, out)
+	if err != nil {
+		return err
+	}
+
+	return nil
```
gtopper, line 273: Just `return api.historyServer.GetLogs(request, out)`?

backends/utils/history_server.go
```diff
@@ -0,0 +1,617 @@
...
+	startTime, err := utils.Str2unixTime(request.Proto.MinStartTime)
+	if err != nil {
+		return errors.Wrap(err, "failed to parse 'start_time' filter")
+	}
+	endTime, err := utils.Str2unixTime(request.Proto.MaxStartTime)
+	if err != nil {
+		return errors.Wrap(err, "failed to parse 'end_time' filter")
+	}
+
+	container, err := m.createV3ioClient(request.Proto.Session, request.Password.Get(), request.Token.Get())
```
gtopper, line 316: Avoid creating per request to avoid go routine leak.

backends/utils/history_server.go
gtopper: Moved the comment to the relevant line. fasthttp isn't the one running go routines in the background. v3io-go does this in the context.

backends/utils/history_server.go
```diff
@@ -0,0 +1,617 @@
...
+	}()
+}
+
+func (m *HistoryServer) createDefaultV3ioClient() {
+
+	session := &frames.Session{}
+
+	token := "99b630dd-bcec-4d3f-a77e-31a3761cef28" // os.Getenv("V3IO_ACCESS_KEY")
+	if token == "" {
+		m.logger.Error("can not create v3io.client. could not find `V3IO_ACCESS_KEY` environment variable")
```
gtopper, line 162: I still think you should return the error and not just log it here.

backends/utils/history_server.go
```diff
@@ -0,0 +1,617 @@
...
+		data = append(data, d...)
+		data = append(data, '\n')
+	}
+
+	logFilePath := m.getCurrentLogFileName()
+	input := &v3io.PutObjectInput{Path: logFilePath, Body: data, Append: true}
+	err := m.container.PutObjectSync(input)
+
+	var retryCount int
+	for err != nil && retryCount < maxRetryNum {
```
gtopper, line 528: Personally I'd just go with infinite retries, but ok I guess.

backends/utils/history_server.go
```diff
@@ -0,0 +1,617 @@
...
+	return frame, nil
+}
+
+func (m *HistoryServer) writeMonitoringBatch(logs []HistoryEntry) {
+	var data []byte
+
+	for _, log := range logs {
+		d, err := json.Marshal(log)
+		if err != nil {
+			m.logger.ErrorWith("Failed to marshal logs to json", "error", err, "log", log)
```
gtopper, line 515: logs -> log

backends/utils/history_server.go
```diff
@@ -0,0 +1,617 @@
...
+			}
+		}
+	}()
+}
+
+func (m *HistoryServer) createDefaultV3ioClient() {
+
+	session := &frames.Session{}
+
+	token := "99b630dd-bcec-4d3f-a77e-31a3761cef28" // os.Getenv("V3IO_ACCESS_KEY")
```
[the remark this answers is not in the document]
talIguaz: 🙄  in the meantime for things to pass i left it. because until platform changes will be pushed no one sets that environment variable.
But ill change it in the PR anyways

backends/utils/history_server.go
talIguaz: hmm i would still rather frames work without logging history, than not work at all. no?

backends/utils/history_server.go
```diff
@@ -0,0 +1,617 @@
...
+	}
+
+	m.HistoryFileNum = defaultHistoryFileNum
+	if m.config.HistoryFileNum != 0 {
+		m.HistoryFileNum = m.config.HistoryFileNum
+	}
+}
+
+func (m *HistoryServer) getCurrentLogFileName() string {
+	// Round current time to the nearest log time-bucket
```
gtopper, line 125: Wrong. It rounds down, not to nearest.

backends/utils/history_server.go
```diff
@@ -0,0 +1,617 @@
...
+	maxLogsInMessage                    = 1000
+	maxRetryNum                         = 5
+
+	queryType   = "query"
+	ingestType  = "ingest"
+	createType  = "create"
+	deleteType  = "delete"
+	executeType = "execute"
+
+	logFileTimeFormat = "2006-01-02T15"
```
gtopper, line 38: You're missing four digits and a Z at the end :)

backends/utils/history_server.go
```diff
@@ -0,0 +1,617 @@
...
+		m.HistoryFileNum = m.config.HistoryFileNum
+	}
+}
+
+func (m *HistoryServer) getCurrentLogFileName() string {
+	// Round current time to the nearest log time-bucket
+	currentTimeInSeconds := time.Now().Unix()
+	currentTimeBucketSeconds := m.HistoryFileDurationSecondSpans * (currentTimeInSeconds / m.HistoryFileDurationSecondSpans)
+	timeBucket := time.Unix(currentTimeBucketSeconds, 0).UTC()
+	return fmt.Sprintf("%v/%v.json", m.LogsFolderPath, timeBucket.Format(logFileTimeFormat))
```
gtopper, line 129: `return m.getLogFileNameByTime(currentTimeBucketSeconds)`?

backends/utils/history_server.go
talIguaz: 💩

backends/utils/history_server.go
gtopper: You can still log it and move on eventually, but the way it's currently written you can't fail it even if you wanted to.

backends/utils/history_server.go
```diff
@@ -0,0 +1,618 @@
...
+	}
+
+	m.HistoryFileNum = defaultHistoryFileNum
+	if m.config.HistoryFileNum != 0 {
+		m.HistoryFileNum = m.config.HistoryFileNum
+	}
+}
+
+func (m *HistoryServer) getCurrentLogFileName() string {
+	// Round current time to the according log time-bucket
```
gtopper, line 126: Round current time down to the nearest time bucket

PR #454, opened 2020-05-07
IG-15649: use num VNs workers in KV write.

gtopper:


backends/kv/writer.go
```diff
@@ -428,14 +414,12 @@ func validColName(name string) string {
 
 // WaitForComplete waits for write to complete
 func (a *Appender) WaitForComplete(timeout time.Duration) error {
-	a.logger.DebugWith("WaitForComplete", "sent", a.sent)
-	a.commChan <- a.sent
-	<-a.doneChan
+	a.Close()
```
talIguaz, line 417: waitForComplete should wait for all items to be sent, this just closes the channel and returns but writes still may happen after that, and you can get errors as well.

backends/kv/writer.go
```diff
@@ -485,18 +485,17 @@ func (a *Appender) updateItemWorker() {
...
+			} else if isOnlyNewItemUpdateModeItemExistError(err, req.UpdateMode) {
 				a.logger.Info("trying to write to an existing item with update mode 'CreateNewItemsOnly' (item: '%v')", req)
 			} else {
-				a.logger.ErrorWith("failed-write response", "error", resp.Error)
-				a.asyncErr = resp.Error
+				a.logger.ErrorWith("failed to update item", "error", err)
+				a.asyncErr = err
 			}
+		} else {
+			resp.Release()
```
talIguaz, line 498: why not always release?

gtopper: Because it will fail badly 🔥. v3io-go releases in case it returns an error.

PR #455, opened 2020-05-12
[DOC] README: Update links to the Iguazio doc site in light of the move to separate versioned doc site

Sharon-iguazio:
`<base URL>/<section>/<version>/` > `<base URL>/<version>/<section>/`
(The old URLs still work because we added redirect rules, but it's better to link to the new URLs.)


Sharon-iguazio:
@dinal please merge.

PR #466, opened 2020-06-09
print version

dinal:


go.mod
```diff
@@ -10,6 +10,7 @@ require (
 	github.com/nuclio/errors v0.0.1
 	github.com/nuclio/logger v0.0.1
 	github.com/nuclio/zap v0.0.2
+	github.com/pavius/impi v0.0.3 // indirect
```
gtopper, line 13: I think this shouldn't be here (go mod tidy)

cmd/framesd/framesd.go
```diff
@@ -45,24 +45,19 @@ var (
...
-	flag.BoolVar(&config.showVersion, "version", false, "show version and exit")
 	flag.Parse()
 
 	log.SetFlags(0) // Show only messages
 
-	if config.showVersion {
-		fmt.Printf("%s version %s\n", path.Base(os.Args[0]), Version)
-		return
-	}
+	fmt.Printf("%s version %s\n", path.Base(os.Args[0]), Version)
```
gtopper, line 60: Just make sure it prints what you expect?

dinal: tried building locally and copiy-tag it to a cluster and got "frames version latest" (which is the name when building locally)
i'm guessing that with an official release tag it will be right

go.mod
dinal: it should, it's used when linting
i pushed go mod tidy commit

gtopper: Doesn't mean it should stay there.

PR #471, opened 2020-06-18
IG-15931

talIguaz:
support passing range-index from framesd + reset_index for history

### DIFF
--- a/column.go
+++ b/column.go
@@ -382,28 +382,32 @@ func NewLabelColumn(name string, value interface{}, size int) (Column, error) {
 		Name: name,
 		Size: int64(size),
 	}
-
-	switch value.(type) {
-	case bool:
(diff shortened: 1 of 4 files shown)


column.go
```diff
@@ -382,28 +382,32 @@ func NewLabelColumn(name string, value interface{}, size int) (Column, error) {
...
-		msg.Ints = []int64{value.(int64)}
-	case string:
-		msg.Dtype = pb.DType_STRING
-		msg.Strings = []string{value.(string)}
-	case time.Time:
-		msg.Dtype = pb.DType_TIME
-		msg.Times = []int64{value.(time.Time).UnixNano()}
-	default:
-		return nil, fmt.Errorf("unknown data type %T", value)
+	if value == nil {
```
gtopper, line 385: What's wrong with `case nil`?

PR #480, opened 2020-06-24
IG-16104

talIguaz:
validate attribute names

### DIFF
--- a/backends/kv/writer_test.go
+++ b/backends/kv/writer_test.go
@@ -100,59 +100,72 @@ func (suite *WriterTestSuite) TestGenExpr() {
 
 func (suite *WriterTestSuite) TestValidateFrameInput() {
 	column, err := frames.NewLabelColumn("col1", 5, 1)
-	suite.NoError(err)
+	suite.Require().NoError(err)
 	columns := []frames.Column{column}
 	frame, err := frames.NewFrame(columns, nil, nil)
-	if err != nil {
-		suite.NoError(err)
-	}
+	suite.Require().NoError(err)
 	writeRequest := frames.WriteRequest{}
 	err = validateFrameInput(frame, &writeRequest)
-	suite.NoError(err)
+	suite.Require().NoError(err)
 }
 
 func (suite *WriterTestSuite) TestValidateFrameInputEmptyColumnName() {
 	column, err := frames.NewLabelColumn("", 5, 1)
-	suite.NoError(err)
+	suite.Require().NoError(err)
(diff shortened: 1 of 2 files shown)


backends/kv/writer_test.go
```diff
@@ -141,6 +141,20 @@ func (suite *WriterTestSuite) TestValidateFrameInputRepeatingShardingKey() {
 	suite.Equal("column 'col1' appears more than once as a partition key", err.Error())
 }
 
+func (suite *WriterTestSuite) TestValidateFrameInputInvalidColumnName() {
```
gtopper, line 144: Also add a positive test? One for name length?

backends/kv/writer_test.go
```diff
@@ -141,6 +141,20 @@ func (suite *WriterTestSuite) TestValidateFrameInputRepeatingShardingKey() {
...
+	column, err := frames.NewLabelColumn("****T-dawg!", 5, 1)
+	suite.NoError(err)
```
gtopper, line 146: `Require()` 😛

backends/kv/writer_test.go
```diff
@@ -141,6 +141,20 @@ func (suite *WriterTestSuite) TestValidateFrameInputRepeatingShardingKey() {
...
+	columns := []frames.Column{column}
+	frame, err := frames.NewFrame(columns, nil, nil)
+	if err != nil {
+		suite.NoError(err)
```
gtopper, line 150: `Require()` 😛 
No need to check `err != nil`

backends/kv/writer_test.go
```diff
@@ -141,6 +141,20 @@ func (suite *WriterTestSuite) TestValidateFrameInputRepeatingShardingKey() {
...
+	}
+	writeRequest := frames.WriteRequest{}
+	err = validateFrameInput(frame, &writeRequest)
+	suite.Error(err)
```
gtopper, line 154: `Require()` 😛

backends/kv/writer_test.go
```diff
@@ -141,6 +141,20 @@ func (suite *WriterTestSuite) TestValidateFrameInputRepeatingShardingKey() {
...
+	suite.Equal("column '****T-dawg!' has an invalid name", err.Error())
```
gtopper, line 155: `Require()` 😛

backends/kv/writer_test.go
talIguaz: Added a test for name length
a positive already exist in that suite

PR #487, opened 2020-06-25
IG-15373: Fail on request args not relevant to request backend.

gtopper:
Implements check for Read and Delete reqeusts for all backends.


backends/backends.go
```diff
@@ -64,3 +67,46 @@ func GetFactory(typ string) Factory {
...
 }
+
+var globalRequestFieldsByRequestType = map[reflect.Type]map[string]bool{
+	reflect.TypeOf(pb.ReadRequest{}): {
+		"Session":      true,
+		"Backend":      true,
+		"Schema":       true,
+		"DataFormat":   true,
+		"RowLayout":    true,
+		"MultiIndex":   true,
```
talIguaz, line 78: MultiIndex, Query, GroupBy are only for TSDB
Join, Marker - are not really supported yet

gtopper: Leaving the concept of unsupported operations out for now, as they will fail in this PR as well. The only consideration is whether we want to make a list of globally unsupported operations to make the error message a bit more helpful.

PR #497, opened 2020-07-01
History error propagate

talIguaz:


api/api.go
```diff
@@ -266,6 +266,9 @@ func (api *API) Exec(request *frames.ExecRequest) (frames.Frame, error) {
 }
 
 func (api *API) History(request *frames.HistoryRequest, out chan frames.Frame) error {
+	if api.historyServer == nil {
+		return errors.New("history server was not initialized properly. restart framesd to access this feature")
```
gtopper, line 270: Can we save the reason it failed to initialize and report it here? Otherwise, this is super annoying for the user (because init will obviously fail again if they restart the service).

gtopper: Also to nitpick: `restart` -> `Restart`, `access` -> `enable`

talIguaz: i log it in the frames pod log. the error will probably say nothing to the user

gtopper: It's much easier to get the error here than have to go search for it in the beginning of the pod log...

gtopper: We decided to tell the user to contact their system administrator.

PR #504, opened 2020-07-02
Framulate - release unneeded data on task complete

talIguaz:


repeatingtask/worker.go
```diff
@@ -56,6 +56,14 @@ func (w *worker) handleTask(task *Task) error {
 		}
 	}
 
+	if task.OnCompleteHandler != nil {
+		err = task.OnCompleteHandler(task.Cookie)
+		task.ErrorsChan <- &TaskError{
+			Repetition: int(repetitionIndex),
+			Error:      err,
```
gtopper, line 63: Hm, I'm pretty sure this should be inside an `if err != nil {`

talIguaz: oopsie

gtopper: Also, why can't we just stick this stuff in the end of the `Handler` function? Why do we need `OnCompleteHandler`?

PR #505, opened 2020-07-02
IG-15749 - infer schema support bucketed tables

talIguaz:


backends/kv/inferschema.go
```diff
@@ -33,8 +34,9 @@ import (
 )
 
 var (
-	intType   = reflect.TypeOf(1)
-	floatType = reflect.TypeOf(1.0)
+	intType            = reflect.TypeOf(1)
+	floatType          = reflect.TypeOf(1.0)
+	hashedBucketFormat = regexp.MustCompile("^[a-zA-z0-9]+_[0-9]+$")
```
gtopper, line 39: `^[a-zA-Z0-9]+_[0-9]+$`, or it will cover a bunch of other random characters between upper and low case on the ascii table.

backends/kv/inferschema.go
```diff
@@ -83,15 +85,23 @@ func schemaFromKeys(keyField string, rowSet []map[string]interface{}) (v3ioutils
 	columnCanBeFullKey := make(map[string]bool)
 	columnCanBePrimaryKey := make(map[string]bool)
 	columnCanBeSortingKey := make(map[string]bool)
+	columnCanBeHashedPrimeryKey := make(map[string]bool)
```
gtopper, line 88: columnCanBeHashedPrimeryKey -> columnCanBeHashedPrimaryKey

backends/kv/inferschema.go
gtopper: Also, shouldn't it be `^[a-zA-Z_][a-zA-Z0-9_]*_[0-9]+$`?

gtopper: In line with #480.

backends/kv/inferschema.go
```diff
@@ -83,15 +85,23 @@ func schemaFromKeys(keyField string, rowSet []map[string]interface{}) (v3ioutils
...
 		var primaryKeyValue string
 		var sortingKeyValue string
+		var hashedPrimaryKeyValue string
 		indexOfDot := strings.Index(keyValue, ".")
 		if indexOfDot >= 0 && indexOfDot < len(keyValue)-1 {
 			sortingKeyValue = keyValue[indexOfDot+1:]
 			primaryKeyValue = keyValue[:indexOfDot]
+
+			if hashedBucketFormat.MatchString(primaryKeyValue) {
+				indexOfUnderscore := strings.Index(primaryKeyValue, "_")
```
gtopper, line 101: Should probably be last index of underscore, not the first one.

backends/kv/inferschema.go
```diff
@@ -83,15 +85,23 @@ func schemaFromKeys(keyField string, rowSet []map[string]interface{}) (v3ioutils
...
 
 	for _, row := range rowSet {
 		keyValue := row["__name"].(string)
 		var primaryKeyValue string
 		var sortingKeyValue string
+		var hashedPrimaryKeyValue string
 		indexOfDot := strings.Index(keyValue, ".")
 		if indexOfDot >= 0 && indexOfDot < len(keyValue)-1 {
 			sortingKeyValue = keyValue[indexOfDot+1:]
 			primaryKeyValue = keyValue[:indexOfDot]
```
gtopper, line 98: Better move this to else than set and unset it...

backends/kv/inferschema.go
```diff
@@ -131,16 +141,30 @@ func schemaFromKeys(keyField string, rowSet []map[string]interface{}) (v3ioutils
 				}
 				columnCanBeSortingKey[attrName] = columnCanBeSortingKey[attrName] && attrValueAsString == sortingKeyValue
 			}
+			if hashedPrimaryKeyValue != "" {
+				if _, ok = columnCanBeHashedPrimeryKey[attrName]; !ok {
+					columnCanBeHashedPrimeryKey[attrName] = true
+				}
+				columnCanBeHashedPrimeryKey[attrName] = columnCanBeHashedPrimeryKey[attrName] && attrValueAsString == hashedPrimaryKeyValue
```
gtopper, line 148: If it's false, set it to true, then use it in a condition? Why not replace these 4 lines with:
```
columnCanBeHashedPrimeryKey[attrName] = attrValueAsString == hashedPrimaryKeyValue
```

talIguaz: **if it does not exists initialize it with true**, you want to have an accumulative indication not just for the current row.

gtopper: Got it. Would be simpler if default was false :)

backends/kv/inferschema.go
```diff
@@ -83,15 +85,24 @@ func schemaFromKeys(keyField string, rowSet []map[string]interface{}) (v3ioutils
...
 		var primaryKeyValue string
 		var sortingKeyValue string
+		var hashedPrimaryKeyValue string
 		indexOfDot := strings.Index(keyValue, ".")
 		if indexOfDot >= 0 && indexOfDot < len(keyValue)-1 {
 			sortingKeyValue = keyValue[indexOfDot+1:]
-			primaryKeyValue = keyValue[:indexOfDot]
+			potentialPrimaryKey := keyValue[:indexOfDot]
+
+			if hashedBucketFormat.MatchString(potentialPrimaryKey) {
```
gtopper, line 100: We should probably not break the case where the keys match the regex by mistake, not because of hashing. The user is still allowed to use `tal_1` and `tal_2` as keys. So a test like this should probably pass (but currently doesn't):
```
func (suite *InferSchemaTestSuite) TestInferSchemaWithHashedPrimaryKey() {
	expectedFields := []v3ioutils.OldSchemaField{
		{Name: "name", Type: "string", Nullable: false},
		{Name: "animal", Type: "string", Nullable: false},
		{Name: "age", Type: "long", Nullable: true},
	}

	keyField := ""
	rowSet := []map[string]interface{}{
		{"__name": "rocky_2.dog", "name": "rocky_2", "animal": "dog", "age": 2},
		{"__name": "mocha_2.dog", "name": "mocha_2", "animal": "dog", "age": 3},
		{"__name": "mocha_4.cat", "name": "mocha_4", "animal": "cat", "age": 9},
	}
	schema, err := schemaFromKeys(keyField, rowSet)
	suite.Require().NoError(err)
	concreteSchema := schema.(*v3ioutils.OldV3ioSchema)
	suite.Require().Equal("name", concreteSchema.Key)
	suite.Require().Equal("animal", concreteSchema.SortingKey)
	suite.Require().ElementsMatch(expectedFields, concreteSchema.Fields)
	suite.Require().Equal(0, concreteSchema.HashingBucketNum)
}
```

PR #507, opened 2020-07-05
IG-15373: Fix erroneous validation error on non-TSDB delete.

gtopper:
Trim down delete request over HTTP.

### DIFF
--- a/clients/py/v3io_frames/http.py
+++ b/clients/py/v3io_frames/http.py
@@ -152,12 +152,17 @@ class Client(ClientBase):
             'session': pb2py(self.session),
             'backend': backend,
             'table': table,
-            'filter': filter,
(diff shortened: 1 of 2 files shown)


clients/py/v3io_frames/http.py
```diff
@@ -152,13 +152,18 @@ def _delete(self, backend, table, filter, start, end, if_missing, metrics):
...
             'backend': backend,
             'table': table,
-            'filter': filter,
-            'start': start,
-            'end': end,
             'if_missing': if_missing,
-            'metrics': metrics,
         }
 
+        if filter:
```
talIguaz, line 158: Just change the default value of `Metrics` in the delete function definition in client.py:342
the filter,start,end change is redundant

PR #512, opened 2020-07-09
IG-16254

talIguaz:
* retry on 409 on history log write
* reties for history read 
* optimize read based on time filter

### DIFF
--- a/backends/utils/history_server.go
+++ b/backends/utils/history_server.go
@@ -3,7 +3,9 @@ package utils
 import (
 	"encoding/json"
 	"fmt"
+	"net/http"
 	"os"
+	"strings"
 	"time"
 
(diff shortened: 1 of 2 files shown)


backends/utils/history_server.go
```diff
@@ -569,6 +580,26 @@ func (m *HistoryServer) StartEvictionTask() {
 	}()
 }
 
+func (m *HistoryServer) filterLogFileByPath(filter historyFilter, path string) (bool, error) {
+	lastSlash := strings.LastIndex(path, "/")
+	logFileName := path[lastSlash+1 : len(path)-5] // removing the `.json` suffix
```
gtopper, line 585: Let's hope it won't find something you don't expect in there...

PR #527, opened 2020-08-25
IG-16686 add timeout in kv write

dinal:


backends/kv/writer.go
```diff
@@ -476,9 +476,20 @@ func validColName(name string) string {
...
+	}
 	close(a.requestChan)
-	<-a.doneChan
-	return a.asyncErr
+	select {
+		case <-a.doneChan:
+			return a.asyncErr
+		case <-time.After(maxWaitTime):
+			err := errors.Errorf("The operation timed out after %.2f seconds.", maxWaitTime.Seconds())
+			return err
```
gtopper, line 491: Better inline

backends/kv/writer.go
```diff
@@ -476,9 +476,20 @@ func validColName(name string) string {
 
 // WaitForComplete waits for write to complete
 func (a *Appender) WaitForComplete(timeout time.Duration) error {
+	var maxWaitTime time.Duration
+	if timeout == 0 {
+		maxWaitTime = 24 * time.Hour
+	} else if timeout > 0 {
+		maxWaitTime = timeout
```
gtopper, line 483: Do we want the hard coded default? Do we want to treat negative values as zero?

gtopper (approved):

PR #541, opened 2020-09-30
handle different timezones in a series

dinal:


clients/py/v3io_frames/pbutils.py
```diff
@@ -227,11 +227,8 @@ def series2col_with_dtype(s, name, dtype):
         kw['bools'] = s
         kw['dtype'] = fpb.BOOLEAN
     elif dtype == fpb.TIME:
-        if s.dt.tz:
-            try:
-                s = s.dt.tz_localize(pytz.UTC)
-            except TypeError:
-                s = s.dt.tz_convert('UTC')
+        s = pd.to_datetime(s, utc=True)
+        s.dt.tz_convert('UTC')
```
gtopper, line 231: Hm, I don't think this line does anything. Should be:
```
s = s.dt.tz_convert('UTC')
```
no?

PR #550, opened 2020-10-25
retry to Add metric if AddFast fails

dinal:


backends/tsdb/writer.go
```diff
@@ -193,7 +193,11 @@ func (a *tsdbAppender) Add(frame frames.Frame) error {
 				} else {
 					err := a.appender.AddFast(metric.ref, tarray[i], values[idx][i])
 					if err != nil {
-						return errors.Wrap(err, "failed to AddFast")
+						//retry with Add in case ref was evicted from cache
+						metric.ref, err = a.appender.Add(metric.lset, tarray[i], values[idx][i])
+						if err != nil {
+							return errors.Wrap(err, "failed to AddFast")
```
gtopper, line 199: At this point we failed to `Add`, didn't we?

backends/tsdb/writer.go
```diff
@@ -193,7 +193,11 @@ func (a *tsdbAppender) Add(frame frames.Frame) error {
 				} else {
 					err := a.appender.AddFast(metric.ref, tarray[i], values[idx][i])
 					if err != nil {
-						return errors.Wrap(err, "failed to AddFast")
+						//retry with Add in case ref was evicted from cache
+						metric.ref, err = a.appender.Add(metric.lset, tarray[i], values[idx][i])
+						if err != nil {
```
gtopper, line 198: Maybe this should be more specific as in https://github.com/v3io/prometheus/pull/73?

gtopper (approved):

PR #556, opened 2020-11-05
IG-17095 check client and server version match

dinal:


api/api.go
```diff
@@ -272,6 +274,11 @@ func (api *API) History(request *frames.HistoryRequest, out chan frames.Frame) e
 	return api.historyServer.GetLogs(request, out)
 }
 
+// Exec executes a command on the backend
```
talIguaz, line 277: wrong comment

cmd/framesd/framesd.go
```diff
@@ -88,7 +88,7 @@ func main() {
 		}
 	}
 
-	hsrv, err := framesHttp.NewServer(cfg, config.httpAddr, framesLogger, historyServer)
+	hsrv, err := framesHttp.NewServer(cfg, config.httpAddr, framesLogger, historyServer, Version)
```
talIguaz, line 91: It doesn't look like `Version` is being set

http/server.go
```diff
@@ -440,6 +440,33 @@ func (s *Server) handleExec(ctx *fasthttp.RequestCtx) {
 	})
 }
 
+func (s *Server) handleVersion(ctx *fasthttp.RequestCtx) {
+	if !ctx.IsPost() { // ctx.PostBody() blocks on GET
+		ctx.Error("unsupported method", http.StatusMethodNotAllowed)
+	}
+
+	requestInner := &pb.VersionRequest{}
```
talIguaz, line 448: you expect an empty request, why parse it and all that?

types.go
```diff
@@ -343,6 +343,12 @@ const (
 	FailOnError = pb.ErrorOptions_FAIL
 )
 
+type VersionRequest struct {
```
talIguaz, line 346: doesn't look like you need any of that

http/server.go
```diff
@@ -61,7 +61,7 @@ type Server struct {
 }
 
 // NewServer creates a new server
-func NewServer(config *frames.Config, addr string, logger logger.Logger, historyServer *utils.HistoryServer) (*Server, error) {
+func NewServer(config *frames.Config, addr string, logger logger.Logger, historyServer *utils.HistoryServer, version string) (*Server, error) {
```
talIguaz, line 64: If you need to have this in both "servers" maybe you can just have this statically here, meaning just dont go through the `api` when requesting version

grpc/server.go
```diff
@@ -348,3 +348,17 @@ func (s *Server) History(request *pb.HistoryRequest, stream pb.Frames_HistorySer
 
 	return apiError
 }
+
+// Exec executes a command
+func (s *Server) Version(ctx context.Context, req *pb.VersionRequest) (*pb.VersionResponse, error) {
+	request := frames.VersionRequest{}
+
+	version, err := s.api.Version(&request)
```
talIguaz, line 356: again, no need to pass anything to the `s.api.Version`
Or alternatively just return `s.version`

clients/py/v3io_frames/grpc.py
```diff
@@ -232,6 +238,17 @@ def do_history(self, backend, container, table, user, action, min_start_time, ma
         for frame in stub.History(request):
             yield msg2df(frame, self.frame_factory)
 
+    @grpc_raise(VersionError)
+    def _check_version(self):
+        stub = fgrpc.FramesStub(self._channel)
+        request = fpb.VersionRequest()
+        resp = stub.Version(request)
+        if resp.version:
```
talIguaz, line 246: is it ok to not get a version?

clients/py/v3io_frames/grpc.py
```diff
@@ -232,6 +238,17 @@ def do_history(self, backend, container, table, user, action, min_start_time, ma
...
+            if __version__ != resp.version:
+                warnings.warn("Warning - Server version \'" + resp.version + "\' is different from client version \'" + __version__ + "\'. Some operations may not work.")
+            else:
```
talIguaz, line 249: the else should be one indent back :)

clients/py/v3io_frames/grpc.py
```diff
@@ -38,6 +40,8 @@
     ('grpc.max_receive_message_length', GRPC_MESSAGE_SIZE),
 ]
 
+warnings.formatwarning = lambda msg, *args, **kwargs: f'{msg}\n'
```
talIguaz, line 43: why are you purposely ignoring args and kwargs?

clients/py/v3io_frames/grpc.py
```diff
@@ -232,6 +238,17 @@ def do_history(self, backend, container, table, user, action, min_start_time, ma
...
             yield msg2df(frame, self.frame_factory)
 
+    @grpc_raise(VersionError)
+    def _check_version(self):
+        stub = fgrpc.FramesStub(self._channel)
+        request = fpb.VersionRequest()
+        resp = stub.Version(request)
+        if resp.version:
+            if __version__ != resp.version:
+                warnings.warn("Warning - Server version \'" + resp.version + "\' is different from client version \'" + __version__ + "\'. Some operations may not work.")
```
talIguaz, line 248: "Some operations may not work as expected" ?

clients/py/v3io_frames/http.py
```diff
@@ -224,6 +230,30 @@ def _history(self, backend, container, table, user, action, min_start_time, max_
...
+        except json.JSONDecodeError as err:
+            raise VersionError(str(err))
+
+        version = out.get('version')
+        if not version:
+            return
+
+        if __version__ != version:
+            warnings.warn("Warning - Server version \'" + version + "\' is different from client version \'" + __version__ + "\'. Some operations may not work.")
+        else:
```
talIguaz, line 254: Looks like this should should be in the `if not version`. no?

cmd/framesd/framesd.go
dinal: it does, by the build process
see in makefile -> FRAMES_BUILD_COMMAND

clients/py/v3io_frames/grpc.py
dinal: warnings prints the msg in a very ugly format - 
`/Users/dinaleventol/PycharmProjects/framestest/v3io_frames_local/http.py:254: UserWarning: Warning - Server version 'unknown' is different from client version '0.4.1'. Some operations may not work.
  __version__ + "\'. Some operations may not work.")` 

googled it a bit and found this way of stripping the extra info warnings prints and keeping just the warn msg itself

talIguaz (approved):

Makefile
```diff
@@ -45,7 +45,7 @@ grpc-go:
 .PHONY: grpc-py
 grpc-py:
 	cd clients/py && \
-	pipenv run python -m grpc_tools.protoc \
+	FRAMES_PYTHON_DEBUG=true pipenv run python -m grpc_tools.protoc \
```
talIguaz, line 48: Do we need this flag here? it's just building the proto

clients/py/v3io_frames/grpc.py
```diff
@@ -87,6 +92,9 @@ def __init__(self, *args, **kwargs):
         # create the session object, persist it between requests
         self._open_new_channel()
 
+        if self._channel and not os.environ.get("FRAMES_PYTHON_DEBUG"):
+            self._check_version()
```
talIguaz, line 96: better put this check in `__init__/Client` which would catch grpc and http

talIguaz (approved):

PR #570, opened 2020-12-03
[POST 3.0] adding infer_schema option to read

katyakats:


backends/kv/reader.go
```diff
@@ -102,13 +102,47 @@ func (kv *Backend) Read(request *frames.ReadRequest) (frames.FrameIterator, erro
...
+				if !request.Proto.InferSchema {
+					return nil, errors.New(
+						fmt.Sprintf("Failed to find a schema for table \"/%s/%s\"; "+
+							"use the `execute` 'infer' command to infer the schema and generate a schema file.",
+							request.Proto.Session.Container, request.Proto.Table))
+				}
+
+				iterInfer, err := v3ioutils.NewAsyncItemsCursor(
+					container, &input, kv.numWorkers, request.Proto.ShardingKeys, kv.logger, 10, partitions,
+					request.Proto.SortKeyRangeStart, request.Proto.SortKeyRangeEnd)
```
dinal, line 114: use the config you added "maxRecordsInfer" instead of 10

katyakats: are you sure? i think i passed 10...

PR #589, opened 2021-01-17
IG-17893: Fix wrapping of pandas-compatible dataframes e.g. cuDF.

gtopper:


clients/py/v3io_frames/client.py
```diff
@@ -299,7 +299,7 @@ def write(self, backend, table, dfs, expression='', condition='',
             Write result
         """
         self._validate_request(backend, table, WriteError)
-        if isinstance(dfs, pd.DataFrame):
+        if type(dfs).__name__ == 'DataFrame':  # This can a pandas.DataFrame or a pandas-compatible DataFrame such as cuDF
```
talIguaz, line 302: Isn't it better to just check for DataFrame/cudf.DataFrame? what other "pandas-compatible" df there are?

gtopper: Dunno. Anyway you can't check for cudf without adding it as a dependency. So it's either name or some kind of attribute.

talIguaz (approved):

PR #617, opened 2021-09-23
IG-19426: fix client.write() when first column is all null

gtopper:


column.go
```diff
@@ -57,6 +57,8 @@ func (c *colImpl) Len() int {
 		return len(c.msg.Times)
 	case pb.DType_BOOLEAN:
 		return len(c.msg.Bools)
+	case pb.DType_NULL:
+		return -1
```
dinal, line 61: it already returned -1 (out of the switch case)

frame.go
```diff
@@ -173,11 +173,7 @@ func (fr *frameImpl) Labels() map[string]interface{} {
 
 // Len is the number of rows
 func (fr *frameImpl) Len() int {
-	if len(fr.columns) > 0 {
-		return fr.columns[0].Len()
-	}
-
-	return 0
+	return fr.indices[0].Len()
```
dinal, line 176: do we always have something in indices that you can access the array without checking it's length?
what about empty df?

dinal:
add a test to cover the bug

column.go
gtopper: I know. I wanted to make it explicit and not a "panic? 🤷‍♂️" thing.

frame.go
gtopper: Every dataframe has an index.

gtopper: Ok, turns out this can happen not because of pandas but because of how we populate the structure internally...

dinal (approved):

PR #648, opened 2023-05-04
Fix deprecation error when using newer numpy versions

gtopper:
Fixes the following error:
```
AttributeError: module 'numpy' has no attribute 'object'.
`np.object` was a deprecated alias for the builtin `object`. To avoid this error in existing code, use `object` by itself. Doing this will not modify any behavior and is safe.
```


assaf758 (approved):
tested to work on both my m2 laptop (where the issue was found) and my centos setup.

PR #651, opened 2023-09-14
IG-22140, IG-22141: Format key values to match python

gtopper:


backends/kv/writer.go
```diff
@@ -685,3 +684,32 @@ func valueToTypedExpressionString(value interface{}) string {
...
 }
+
+func valueToKeyString(value interface{}) string {
+	switch typedVal := value.(type) {
+	case string:
+		return typedVal
+	case time.Time:
+		seconds := typedVal.Unix()
+		nanos := typedVal.Nanosecond()
+		return fmt.Sprintf("%v:%v", seconds, nanos)
```
gtopper: Should probably remove this
```suggestion
```

clients/py/dev-requirements.txt
```diff
@@ -1,4 +1,6 @@
 build
+# https://stackoverflow.com/questions/73929564/entrypoints-object-has-no-attribute-get-digital-ocean
```
alxtkr77, line 2: How this is related?

backends/kv/writer.go
```diff
@@ -685,3 +684,32 @@ func valueToTypedExpressionString(value interface{}) string {
...
+		// Python/storey compatibility by capitalizing
+		// IG-22141
+		if typedVal {
+			return "True"
+		}
+		return "False"
+	case float64, float32:
+		// Python/storey compatibility by deleting trailing zeros (except one trailing zero after period)
+		// IG-22140
+		str := fmt.Sprintf("%f", typedVal)
```
alxtkr77, line 702: Could be just str="{:.1f}".format(typedVal) if typedVal.is_integer() else str(typedVal)

backends/kv/writer.go
```diff
@@ -363,13 +363,12 @@ func (a *Appender) Add(frame frames.Frame) error {
 }
 
 func (a *Appender) formatKeyName(key interface{}, sortingVal interface{}) string {
-	var format string
```
alxtkr77, line 366: The change is lacking tests

backends/kv/writer.go
alxtkr77: Why ? Looks like it fixes IG-22141 timestamps. Am I missing something?

gtopper: It's the wrong format – i.e. it's incompatible with storey, which writes timestamps as some version of ISO 8601, pandas that does roughly (but probably not exactly) the same, and spark that hopefully does what pandas does. The difficulty in nailing down a timestamp format is a strong argument in favour of blocking this feature (otherwise, we need to put in the time to standardize this across engines).

clients/py/dev-requirements.txt
gtopper: It's not. It's needed to unbreak development CI.

backends/kv/writer.go
gtopper: This is Go code, not python...

backends/kv/writer.go
gtopper: Added regression tests.

backends/kv/writer.go
gtopper: See comments on [IG-22141](https://jira.iguazeng.com/browse/IG-22141) for discussion.

backends/kv/writer.go
alxtkr77: Its python . Maybe its go as well 😀

gtopper: Well, let me know if you have a suggestion in Go :)

alxtkr77: str := fmt.Sprintf("%.1f", typedVal) if typedVal == float64(int64(typedVal)) else strconv.FormatFloat(typedVal, 'f', -1, 64)

Just for fun :)

alxtkr77 (approved):

assaf758 (approved):

